PR #32, opened 2025-04-17
🧹Front/clean up

jbpenrath:
## Purpose

Let's do some cleaning in the frontend codebase

## Proposal

- [x] Do not ignore Typescript errors at build
- [x] Fix all TS and lint errors
- [x] Setup frontend test environment and write first test suite
- [x] Add CI jobs for frontend healthcheck (lint, test and build)

### DIFF
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -3,6 +3,10 @@
   "version": "0.1.0",
   "private": true,
   "type": "module",
+  "engines": {
+    "node": ">=22.0.0 <23.0.0",
+    "npm": ">=10.0.0"
+  },
   "scripts": {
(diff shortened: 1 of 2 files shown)


src/frontend/package.json
```diff
@@ -2,46 +2,57 @@
...
   "version": "0.1.0",
   "private": true,
+  "type": "module",
+  "engines": {
+    "node": ">=22.0.0 <23.0.0",
+    "npm": ">=10.0.0"
+  },
   "scripts": {
     "dev": "next dev",
     "build": "next build --no-lint",
```
sylvinus: we should remove the `--no-lint` here (I added it as a hack)

PR #36, opened 2025-04-19
🚧🚧🚧 ✨(api) allow to create new message

sdemagny:
Complete permissions required implementation.
Allow to reply (not yet with last message content)
TODO: 

- [ ] send message for real :D

### DIFF
--- a/src/backend/core/api/permissions.py
+++ b/src/backend/core/api/permissions.py
@@ -4,7 +4,7 @@ from django.core import exceptions
 
 from rest_framework import permissions
 
-from core import models
+from core import enums, models
 
 ACTION_FOR_METHOD_TO_PERMISSION = {
     "versions_detail": {"DELETE": "versions_destroy", "GET": "versions_retrieve"},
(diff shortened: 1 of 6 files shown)


src/backend/core/api/permissions.py
```diff
@@ -84,17 +84,31 @@ class IsAllowedToAccessMailbox(IsAuthenticated):
...
         if not mailbox_id and not thread_id:
             return False
+        permissions_required = [enums.MailboxPermissionChoices.READ]
+        if request.method == "POST":
+            permissions_required = [
+                enums.MailboxPermissionChoices.SEND,
+                enums.MailboxPermissionChoices.EDIT,
+            ]
         if mailbox_id:
             return models.Mailbox.objects.filter(
```
sylvinus: seems dangerous to return early. What if user passes mailbox_id that he owns, and an unrelated thread_id ? The has_permission will return true but the other method may only look at threads.

sdemagny: Yes, you're right. I'll rework that part.
Thanks

src/backend/core/api/permissions.py
```diff
@@ -131,3 +131,37 @@ def has_object_permission(self, request, view, obj):
...
+            return False
+        # required permissions to send a message
+        permissions_required = [
+            enums.MailboxPermissionChoices.SEND,
+            enums.MailboxPermissionChoices.EDIT,
+        ]
+        # check if user has access required to send a message with this mailbox
+        return mailbox.accesses.filter(
+            user=request.user,
+            permission__in=permissions_required,
```
sylvinus, line 166: with this code it looks like someone with only EDIT can send ?

src/backend/core/api/viewsets.py
```diff
@@ -318,3 +320,106 @@ def get_queryset(self):
...
+        )
+        if not sender_id:
+            raise drf.exceptions.ValidationError("Either a senderId must be provided.")
+        try:
+            sender = models.Contact.objects.get(id=sender_id)
+        except models.Contact.DoesNotExist as exc:
+            raise drf.exceptions.ValidationError("Sender does not exist.") from exc
+        # get mailbox from sender email split by '@'
+        try:
+            mailbox = models.Mailbox.objects.get(
```
sylvinus: isn't it better to get the mailbox from the `permission_classes` so that we only fetch it once?

sdemagny: good idea!

PR #42, opened 2025-04-21
API to mark message as read and update read status of thread 

sdemagny:
Remove is_read field on Message
Add is_read field on Thread
Add API endpoints to mark message as read or unread


src/backend/core/api/viewsets.py
```diff
@@ -309,17 +310,74 @@ def get_queryset(self):
...
+                return queryset.none()
+        return queryset
+
+    @drf.decorators.action(
+        detail=True,
+        methods=["patch"],
+        url_path="mark-as-read",
+        url_name="mark-as-read",
+    )
+    def mark_as_read(self, request, *args, **kwargs):
```
sylvinus: can we use only one method with a parameter ?

PR #141, opened 2025-06-11
👽️(api) fix openapi schema for import message

sdemagny:
Fix and re-generate schema
Use a celery task to import eml too

### DIFF
--- a/src/backend/core/api/viewsets/import_message.py
+++ b/src/backend/core/api/viewsets/import_message.py
@@ -33,23 +33,19 @@ class ImportViewSet(viewsets.ViewSet):
     @extend_schema(
         request=ImportFileSerializer,
         responses={
-            200: OpenApiResponse(
-                description="EML file import completed successfully",
-                response={
-                    "type": "object",
-                    "properties": {
-                        "task_id": {"type": "string", "description": "Task ID for tracking the import"},
-                        "type": {"type": "string", "description": "Type of import (eml)"},
-                    },
-                },
-            ),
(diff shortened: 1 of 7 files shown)


src/backend/core/api/viewsets/import_message.py
```diff
@@ -15,19 +15,63 @@
...
         responses={
-            202: OpenApiResponse(description="Import started. Returns Celery task ID.")
+            200: OpenApiResponse(
+                description="EML file import completed successfully",
+                response=ImportFileSerializer,
+            ),
+            202: OpenApiResponse(
+                description="MBOX file import started. Returns Celery task ID for tracking.",
+                response=ImportFileSerializer,
+            ),
```
jbpenrath: Hmmm, currently this is not the response in this case.

I get a response with the `task_id`
```json
{"task_id":"57ac9767-7c5c-4139-aac7-c1c2c2cb6158","type":"mbox"}
```

src/backend/core/api/viewsets/import_message.py
```diff
@@ -58,11 +102,93 @@ def import_file(self, request):
         request=ImportIMAPSerializer,
         responses={
             202: OpenApiResponse(
-                description="IMAP import started. Returns Celery task ID."
-            )
+                description="IMAP import started. Returns Celery task ID for tracking the import progress.",
+                response=ImportIMAPSerializer,
+            ),
```
jbpenrath: Ditto, in this case I get a response with the `task_id`

PR #156, opened 2025-06-17
✨(frontend): display labels on thread list view

sdemagny:
- Added LabelBadge component to render labels as colored badges in the thread list and thread view.
- Updated Thread and Label models to support label objects and color.
- Styled label badges and thread item label container for better visual integration.
- Integrated label display in thread list


src/frontend/src/features/layouts/components/label-badge/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+                style={{
+                    backgroundColor: label.color,
+                    color: getContrastColor(label.color),
+                    padding: size === "small" ? "0.125rem 0.375rem" : "0.25rem 0.5rem",
+                    borderRadius: "4px",
+                    fontSize: size === "small" ? "0.75rem" : "0.875rem",
+                    fontWeight: 600,
+                    display: "inline-block",
+                }}
+            >
```
jbpenrath, line 22: Can you juste create a sass file instead of putting that as inline style

src/frontend/src/features/layouts/components/label-badge/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+    const r = parseInt(hex.substring(0, 2), 16)
+    const g = parseInt(hex.substring(2, 4), 16)
+    const b = parseInt(hex.substring(4, 6), 16)
+    
+    // Calculate relative luminance
+    const luminance = (0.299 * r + 0.587 * g + 0.114 * b) / 255
+    
+    // Return black for light colors, white for dark colors
+    return luminance > 0.5 ? "#000000" : "#FFFFFF"
+}
```
jbpenrath:
```suggestion
// Helper function to determine if text should be black or white based on background color
// TODO: Move it in a color (in ui-kit?)
function getContrastColor(hexColor: string): string {
    // Remove the # if present
    const hex = hexColor.replace("#", "")
    
    // Convert to RGB
    const r = parseInt(hex.substring(0, 2), 16)
    const g = parseInt(hex.substring(2, 4), 16)
    const b = parseInt(hex.substring(4, 6), 16)
    
    // Calculate relative luminance
    const luminance = (0.299 * r + 0.587 * g + 0.114 * b) / 255
    
    // Return black for light colors, white for dark colors
    return luminance > 0.5 ? "#000000" : "#FFFFFF"
}
```

Issue #185, opened 2025-06-26
Scalable email storage


sylvinus:
Right now we've taken the shortcut of storing `raw_mime` as a binaryfield in the Message model.

Step 1 should be to make it a Blob instance instead.

Step 2 should be to allow Blob instances to be stored in configurable S3-compatible object storage.

Issue #177, opened 2025-06-26
DNS Records


sylvinus:
The DNS records to configure should be correctly displayed in the admin. Currently WIP.

We should be able to test their existence.

We should be able, in some cases, to automatically apply them. Currently for us that means supporting the Scaleway API at least.

We need to properly spec DKIM key management before this (currently there is only one, global to the server)

PR #265, opened 2025-07-10
✨(all) refactor several areas before 0.1 release

sylvinus:
We are going to release a 0.1 version soon, along with our first production deployment.

Starting from there, migrations and a consistent developer experience will be officially supported.

To make that easier, this much-too-large PR cleans up several areas:

- Reset migrations one last time
- Update models for storage efficiency (move textchoices to integerchoices on high-volume tables)
- Use Blobs for mail mime data and draft bodies. Having them in a separate PG table is a first step, we will later start offloading them to object storage.
- Add default ZSTD compression to blobs
- Add per-domain DKIM Keys
- Add DNS check and provisioning, with a first Scaleway provider
- Fix Keycloak user provisioning
- Fix Attachment storage, they are now stored individually only at the drafting stage. Afterwards they are extracted from the main blob. This may be optimized later but at least we only store once. For JMAP compatibility, this requires using fake IDs in the blob API route.
- Add a management command and recurring task to retry unsent messages
- Improve the local developer experience with new ports and `make` commands
- Repackage MTA-in and MTA-out to be closer to Backend: Poetry, multi-step Dockerfile, move compose and makefile to the root
- Migrate to OpenSearch
- Improve overall documentation and add a self-hosting page

Contributes to #177 and #185

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,15 +130,15 @@ down: ## stop and remove containers, networks, images, and volumes
 	@$(COMPOSE) down
 .PHONY: down
 
-logs: ## display backend-dev logs (follow mode)
-	@$(COMPOSE) logs -f backend-dev
+logs: ## display all services logs (follow mode)
+	@$(COMPOSE) logs -f
 .PHONY: logs
 
-start: ## start full stack
+start: ## start all development services
 	@$(COMPOSE) up --force-recreate --build -d frontend-dev backend-dev celery-dev mta-in
 .PHONY: start
 
-start-minimal: ## start minimal stack: only backend, frontend, keycloak and DB
+start-minimal: ## start minimal services (backend, frontend, keycloak and DB)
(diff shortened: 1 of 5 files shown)


src/backend/core/api/serializers.py
```diff
@@ -134,6 +134,7 @@ class BlobSerializer(serializers.ModelSerializer):
     """Serialize blobs."""
 
     blobId = serializers.UUIDField(source="id", read_only=True)
+    type = serializers.CharField(source="content_type", read_only=True)
```
sdemagny, line 203: le champ maintenant c'est content_type, tu peux supprimer cette ligne et changer dans la liste des fields plus bas ligne 143 ou 144 "type" -> "content_type"

sylvinus: ah, c'était pour pas le changer côté API

sylvinus: (JMAP utilise "type")

PR #284, opened 2025-07-20
✨(admin) improve Admin mailbox management page

sylvinus:
This fixes a few small anoyances and adds a new field Mailbox.is_identity to the models so that we can save what type of mailbox the admin created.


src/backend/core/api/viewsets/maildomain.py
```diff
@@ -202,12 +212,40 @@ def create(self, request, *args, **kwargs):
...
-        if mailbox_type == "personal" and settings.IDENTITY_PROVIDER == "keycloak":
+
+        # This is a somewhat hacky bypass of abstractions, but for now
+        # we need to return a one time password synchronously
+        if (
+            mailbox_type == "personal"
+            and settings.IDENTITY_PROVIDER == "keycloak"
+            and is_identity
+            and domain.identity_sync
+        ):
```
jbpenrath: Checking `mailbox_type` value and `is_identity` is redundant

```suggestion
        if (
            is_identity
            and settings.IDENTITY_PROVIDER == "keycloak"
            and domain.identity_sync
        ):
```

src/frontend/src/features/layouts/components/admin/modal-create-address/index.tsx
```diff
@@ -131,6 +134,16 @@ export const ModalCreateAddress = ({ isOpen, onClose }: ModalCreateAddressProps)
...
 
+  // Focus first field when tab changes
+  useEffect(() => {
+    if (firstFieldRef && isOpen) {
+      // Small delay to ensure the field is rendered
+      setTimeout(() => {
+        firstFieldRef.focus();
+      }, 100);
+    }
+  }, [activeTab, isOpen, firstFieldRef]);
```
jbpenrath, line 146: Does it work like that ? In this way you should be able to remove the field ref and the hacky timeout.
```suggestion
  useEffect(() => {
    if (activeTab === 'personal') form.setFocus('first_name');
    else if (activeTab === 'shared') form.setFocus('name');
    else if (activeTab === 'redirect') form.setFocus('prefix');
  }, [form.setFocus, activeTab]);
```

jbpenrath (approved):

src/frontend/src/features/layouts/components/admin/modal-create-address/index.tsx
sylvinus: Sorry can't make it to work :/ Even from console form.setFocus() doesn't seem to have any effect

jbpenrath: Ok we gave it a try : /

PR #286, opened 2025-07-21
✨(ai): add ai summarization for threads

MathisRouget:
## Purpose

Some threads and messages can be very long. Summarizing them helps users quickly retrieve key information.  
We also want to assist users in avoiding common mistakes when sending messages.

## Proposal

Introduce an AI-powered summarization feature for threads.  
Also add a simple rule-based system to detect forgotten attachments.

### AI Summarization

- Uses the `AlbertAPI` with the `albert-large` model.
- If a thread is too long, a summary is automatically generated upon reception.
- If no API key is configured, the feature is hidden from the frontend.
- Users can manually update the generated summary.

### Attachment Reminder (non-AI)

- Detects when users mention an attachment (e.g. “see attached”, “voir PJ”) but forget to include a file.
- Displays a warning before sending the message.


env.d/development/backend.defaults
```diff
@@ -71,4 +71,9 @@ MTA_OUT_HOST=mta-out:587
 MTA_OUT_SMTP_USERNAME=user
 MTA_OUT_SMTP_PASSWORD=pass
 MDA_API_SECRET=my-shared-secret-mda
-SALT_KEY=ThisIsAnExampleSaltForDevPurposeOnly
\ No newline at end of file
+SALT_KEY=ThisIsAnExampleSaltForDevPurposeOnly
+
+# AI 
+AI_BASE_URL=https://albert.api.etalab.gouv.fr/v1
```
sylvinus: With this default config, users will have an error by default? It should probably be empty instead

src/backend/core/ai/thread_summarizer.py
```diff
@@ -0,0 +1,29 @@
...
+    Returns:
+        str: Summary of the thread.
+    """
+
+    # Extract messages from the thread
+    messages = get_messages_from_thread(thread)
+
+    # Prepare the prompt for the AI model
+    conversation_text = "\n\n".join([str(message) for message in messages])
+    prompt_query = "Tu es un assistant intelligent qui résume des boucles de mails. Tu dois résumer le contenu de la conversation en une ou deux lignes maximum sans préciser 'Résumé:'. Si des liens importants apparaissent dans les emails, tu dois les mentionner dans le résumé de façon claire en Markdown et intégré au résumé et sinon ne pas les mentionner."
```
sylvinus: We need to store these prompts outside of the code (in a template directory?) and to translate them

src/backend/core/api/viewsets/config.py
```diff
@@ -74,4 +74,17 @@ def get(self, request):
             if hasattr(settings, setting):
                 dict_settings[setting] = getattr(settings, setting)
 
+        # AI_ENABLED logic: True iff all required AI settings are set and not default/empty
+        ai_enabled = False
+        ai_key = getattr(settings, "AI_API_KEY", None)
+        ai_base_url = getattr(settings, "AI_BASE_URL", None)
+        ai_model = getattr(settings, "AI_MODEL", None)
+        if (
+            ai_key not in (None, "", "default", "<my-ai-api-key>")
```
sylvinus: the test should be that there is a non-null AI_BASE_URL

src/backend/core/api/viewsets/thread.py
```diff
@@ -410,6 +411,83 @@ def list(self, request, *args, **kwargs):
...
+    @extend_schema(
+        responses={
+            200: OpenApiResponse(
+                response={
+                    "type": "object",
+                    "properties": {
+                        "summary": {"type": "string"},
+                    },
+                },
+                description="Thread summary retrieved successfully.",
```
sylvinus, line 423: Weird description

src/backend/core/mda/inbound.py
```diff
@@ -574,6 +576,17 @@ def deliver_inbound_message(  # pylint: disable=too-many-branches, too-many-stat
             thread.snippet = new_snippet
             thread.save(update_fields=["snippet"])
 
+        # Update summary if needed
+        messages = get_messages_from_thread(thread)
```
sylvinus: There should be a test to check if AI is enabled here

src/backend/core/ai/utils.py
```diff
@@ -0,0 +1,116 @@
+import uuid
+from dataclasses import dataclass, field
+from datetime import datetime
+from typing import Any, Dict, List
+
+from core.models import Thread
+
+
+@dataclass
+class Message:
```
sylvinus: It feels weird to create a second Message model here. If what you basically need is a message serializer it should be written like this, or even be integrated in the mail model

sylvinus (changes requested):
Great work!

I focused mainly on the backend

We should add `AI_FEATURE_THREAD_SUMMARY=1" env var to enable specific features individually ?

src/backend/core/ai/utils.py
```diff
@@ -0,0 +1,38 @@
...
+def is_ai_enabled() -> bool:
+    """Check if AI features are enabled based on the presence of required settings in the environment (API_KEY, BASE_URL, MODEL)"""
+    ai_key = getattr(settings, "AI_API_KEY", None)
+    ai_base_url = getattr(settings, "AI_BASE_URL", None)
+    ai_model = getattr(settings, "AI_MODEL", None)
+    return (
+        ai_base_url not in (None, "", "default")
+        and ai_model not in (None, "", "default")
+        and ai_key not in (None, "", "default", "<my-ai-api-key>")
+    )
```
jbpenrath: IMO this is not the job of this util to check that developer has provided valid values.
```suggestion
    return (
        ai_base_url is not None
        and ai_model is not None
        and ai_key is not None
    )
```

src/backend/core/ai/utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def is_ai_summary_enabled() -> bool:
+    """
+    Check if AI summary features are enabled.
+    This is determined by the presence of the AI settings and if AI_FEATURE_SUMMARY_ENABLED is set to 1.
+    """
+    return is_ai_enabled() and getattr(settings, "AI_FEATURE_SUMMARY_ENABLED", 0) == 1
```
jbpenrath: `AI_FEATURE_SUMMARY_ENABLED` is a BooleanValue so let's use a boolean statement.


```suggestion
    return is_ai_enabled() and getattr(settings, "AI_FEATURE_SUMMARY_ENABLED", False) is True
```

src/backend/core/ai/thread_summarizer.py
```diff
@@ -0,0 +1,34 @@
...
+        prompts = json.load(f)
+    
+    # Get the appropriate prompt template based on the language
+    prompt_template = prompts.get(language, prompts["fr"])  # Fallback to 'fr' if not found
+    prompt_query = prompt_template["summary_query"]
+    prompt = prompt_query.format(messages=messages_as_text, language=language)
+
+    # Make the API call to get the summary
+    summary = AIService().call_ai_api(prompt)
+    return summary
```
jbpenrath: You could use the i18n features of Django to get the prompt according to the active language through `gettext_lazy` and the `makemessages` command.

Then you could enforce language if needed through `override` method.
In this way, the default language would be the the `LANGUAGE_CODE` settings and the fallback logic will be available for free.

src/backend/core/models.py
```diff
@@ -1023,6 +1024,43 @@ def get_all_recipient_contacts(self) -> Dict[str, List[Contact]]:
...
+        # Sender: "Name <email>" or just email
+        sender = str(self.sender)
+        # Recipients: list of "Name <email>" or just email
+        to_contacts = self.recipients.filter(type=MessageRecipientTypeChoices.TO).select_related("contact")
+        recipients = [str(mr.contact) for mr in to_contacts]
+        # CC
+        cc_contacts = self.recipients.filter(type=MessageRecipientTypeChoices.CC).select_related("contact")
+        cc = [str(mr.contact) for mr in cc_contacts]
+        # Subject
+        subject = self.subject or "Aucun objet"
```
jbpenrath: Should be internationalized.


```suggestion
        from django.utils.translation import gettext_lazy as _
        ...
        subject = self.subject or _("No subject")
```

src/backend/core/api/serializers.py
```diff
@@ -404,6 +404,10 @@ def get_labels(self, instance):
         ).distinct()
         return ThreadLabelSerializer(labels, many=True).data
 
+    def get_summary(self, instance):
+        """Get the summary of the thread."""
+        return instance.summary or ""
+
```
jbpenrath: You could use a `CharField` instead.

```python
class ThreadSerializer(serializers.ModelSerializer):
    """Serialize threads."""
    ...
    summary = serializers.CharField(read_only=True)
    ...
```

src/backend/core/api/viewsets/config.py
```diff
@@ -74,4 +74,17 @@ def get(self, request):
...
+            and ai_base_url not in (None, "", "default")
+            and ai_model not in (None, "", "default")
+        ):
```
jbpenrath: Should be replaced by `is_ai_enabled` utils nope?

src/backend/core/api/viewsets/thread.py
```diff
@@ -410,6 +411,83 @@ def list(self, request, *args, **kwargs):
...
+    @drf.decorators.action(detail=True, methods=["get"], url_path="summary")
+    def get_summary(self, request, id=None):  # pylint: disable=unused-argument
+        """Retrieve the summary of a thread."""
+        thread = self.get_object()
+
+        # Check access rights
+        if not models.ThreadAccess.objects.filter(
+            thread=thread, mailbox__accesses__user=request.user
+        ).exists():
+            return drf.response.Response({"detail": "Permission denied"}, status=403)
```
jbpenrath: No need to do that here. The `get_object` method calls the `get_queryset` method under the hood that is already doing this check.

src/backend/core/api/viewsets/thread.py
```diff
@@ -410,6 +411,83 @@ def list(self, request, *args, **kwargs):
...
+
+        if not thread.summary:
+            thread.summary = ""
+            thread.save()
+
```
jbpenrath: Why do you need to do that?

src/backend/core/api/viewsets/thread.py
```diff
@@ -410,6 +411,83 @@ def list(self, request, *args, **kwargs):
...
+        thread = self.get_object()
+
+        # Check access rights
+        if not models.ThreadAccess.objects.filter(
+            thread=thread, mailbox__accesses__user=request.user
+        ).exists():
+            return drf.response.Response(
+                {"detail": "Permission denied"},
+                status=status.HTTP_403_FORBIDDEN,
+            )
```
jbpenrath: Ditto

src/backend/core/models.py
```diff
@@ -1023,6 +1024,43 @@ def get_all_recipient_contacts(self) -> Dict[str, List[Contact]]:
...
+        msg_id = str(self.id)
+        return (
+            f"Message ID: {msg_id}\n"
+            f"De: {sender}\n"
+            f"À: {', '.join(recipients)}\n"
+            f"CC: {', '.join(cc)}\n"
+            f"Date: {date_str}\n"
+            f"Sujet: {subject}\n\n"
+            f"{body}"
+        )
```
jbpenrath: Ditto. Should be internationalized.

src/backend/core/services/ai_service.py
```diff
@@ -0,0 +1,32 @@
...
+            "messages": [{"role": "user", "content": prompt}],
+            "stream": False,
+            "n": 1,
+        }
+
+        response = self.client.chat.completions.create(**data)
+        content = response.choices[0].message.content
+
+        if not content:
+            raise RuntimeError("AI response does not contain an answer")
```
jbpenrath: What do you think to use `ValueError` instead?
```suggestion
            raise ValueError("AI response does not contain an answer")
```

src/frontend/src/features/forms/components/message-form/index.tsx
```diff
@@ -500,6 +533,12 @@ export const MessageForm = ({
 
                 <AttachmentUploader initialAttachments={getDefaultAttachments()} onChange={form.handleSubmit(saveDraft)} />
 
+                {showAttachmentsForgetAlert && 
+                  <section className="attachments-forget__alert">
+                    <p> {t("attachments.forgot_question")} </p>
+                  </section>
```
jbpenrath:
```suggestion
                  <Banner type="warning">
                    {t("attachments.forgot_question")}
                  </section>
```

src/frontend/src/features/forms/components/message-form/index.tsx
```diff
@@ -358,6 +360,27 @@ export const MessageForm = ({
...
+
+
+    const areAttachmentsMentionnedInDraft = (): boolean => {
+    const messageEditorDraft =
+      form.getValues("messageEditorDraft")?.toLowerCase() || "";
+    return keyWordsAttachments.some((keyword) =>
+      messageEditorDraft.includes(keyword)
+    );
+  };
+
```
jbpenrath: Can you put this check into the `MailHelper` ?

src/frontend/src/features/forms/components/message-form/index.tsx
```diff
@@ -420,6 +443,16 @@ export const MessageForm = ({
...
+    useEffect(() => {
+      const subscription = form.watch((values) => {
+        setShowAttachmentsForgetAlert(
+          areAttachmentsMentionnedInDraft() &&
+          values.attachments?.length === 0
+        );
+      });
+      return () => subscription.unsubscribe();
+    }, [areAttachmentsMentionnedInDraft, form.watch]);
+
```
jbpenrath: Instead of a `useEffect` and `useState`, I feel you could use a `useMemo` to achieve that. That helps to save some renders.

src/frontend/src/features/i18n/translations.json
```diff
@@ -286,6 +288,13 @@
       "message_body": {
         "show_embedded_message": "Show embedded message"
       },
+      "summary": {
+        "title": "Summary",
+        "generation_in_progress": "Summary generation in progress...",
```
jbpenrath:
```suggestion
        "generation_in_progress": "Generating summary...",
```

src/frontend/src/features/i18n/translations.json
```diff
@@ -286,6 +288,13 @@
...
+        "no_summary": "No summary available.",
+        "refresh_success": "Summary refreshed successfully.",
```
jbpenrath:
```suggestion
        "refresh_success": "Summary refreshed!",
```

src/frontend/src/features/i18n/translations.json
```diff
@@ -702,6 +713,13 @@
       "message_body": {
         "show_embedded_message": "Afficher le message inclus"
       },
+      "summary": {
+        "title": "Résumé",
+        "generation_in_progress": "Génération du résumé en cours...",
+        "no_summary": "Aucun résumé disponible.",
+        "refresh_success": "Résumé mis à jour avec succès.",
```
jbpenrath:
```suggestion
        "refresh_success": "Résumé mis à jour !",
```

src/frontend/src/features/layouts/components/thread-view/components/thread-summary/_index.scss
```diff
@@ -0,0 +1,25 @@
...
+  margin: 0;
+}
+
+.thread-summary__refresh-button {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: center;
+  width: 10%;
+  margin: 0;
```
jbpenrath: Should to the job.

```suggestion
.thread-summary__content {
  display: flex;
  flex-direction: column;
  justify-content: center;
  flex: 1;
  min-height: 40px;
  padding-left: var(--c--theme--spacings--base);
  margin: 0;
}

.thread-summary__refresh-button {
  display: flex;
  flex-direction: row;
  align-items: center;
  justify-content: center;
  margin: 0;
}
```

src/frontend/src/features/layouts/components/thread-view/components/thread-summary/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+            ? (data.data as { summary?: string }).summary
+            : undefined;
+
+        if (newSummary) {
+          setLocalSummary(newSummary);
+          cacheNewSummary(newSummary);
+          onSummaryUpdated?.(newSummary);
+          addToast(<ToasterItem type="info">{t("summary.refresh_success")}</ToasterItem>);
+        }
+      },
```
jbpenrath: IMO if you have to do that that seems there is a type issue about `data` argument.

src/frontend/src/features/layouts/components/thread-view/components/thread-summary/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+    },
+  });
+
+  const handleRefresh = () => {
+    refreshMutation.mutate({ id: threadId });
+  };
+
+  useEffect(() => {
+    setLocalSummary(summary);
+  }, [summary]);
```
jbpenrath: Is it really useful ?

At mount, the `useState` is hydrated with the `summary` prop value. Then this value should be always up to date when the mutation is called nope ?

src/frontend/src/features/layouts/components/thread-view/index.tsx
```diff
@@ -74,6 +81,10 @@ export const ThreadView = () => {
     const trashedMessageIds = messages?.results?.filter((m) => m.is_trashed).map((m) => m.id) || [];
     const isThreadTrashed = trashedMessageIds.length === messages?.results?.length;
 
+    // Get AI Configuration
+    const { data: configData } = useConfigRetrieve();
+    const config = configData?.data as ConfigWithAI | undefined;
+    const isAIEnabled = config?.AI_ENABLED === true && config?.AI_FEATURE_SUMMARY_ENABLED === true;
```
jbpenrath: You could put that check in a utils to prevent to repeat this logic across the app.


```suggestion
    // Get AI Configuration
    const { data: configData } = useConfigRetrieve();
    const isAIEnabled = config?.AI_ENABLED === true && config?.AI_FEATURE_SUMMARY_ENABLED === true;
```

src/backend/core/models.py
```diff
@@ -1023,6 +1024,43 @@ def get_all_recipient_contacts(self) -> Dict[str, List[Contact]]:
...
+
+    def get_tokens_count(self) -> int:
+        """Get the number of tokens in the message (word count of get_as_text)."""
+        return len(self.get_as_text().split())
```
sylvinus: We should only count the tokens in the title+body there?

src/backend/core/tests/api/test_config.py
```diff
@@ -33,11 +33,12 @@ def test_api_config(is_authenticated):
...
-    }
+    data = response.json()
+    assert data["ENVIRONMENT"] == "test"
+    assert data["LANGUAGES"] == [["en-us", "English"], ["fr-fr", "French"], ["de-de", "German"]]
+    assert data["LANGUAGE_CODE"] == "en-us"
+    assert data["POSTHOG_KEY"] == "132456"
+    assert data["POSTHOG_HOST"] == "https://test.i.posthog-test.com"
+    assert data["POSTHOG_SURVEY_ID"] == "7890"
+    assert type(data["AI_ENABLED"]) is bool
+    assert type(data["AI_FEATURE_SUMMARY_ENABLED"]) is bool
```
sylvinus: you should leave it like it was, it's a good safeguard against adding unexpected configs in the public scope

env.d/development/backend.defaults
sylvinus: Not sure this is resolved, IMHO the values here should be commented, otherwise users will get errors with a fresh git clone.

src/backend/core/migrations/0001_initial.py
```diff
@@ -55,6 +55,7 @@ class Migration(migrations.Migration):
                 ('has_active', models.BooleanField(default=False, verbose_name='has active')),
                 ('messaged_at', models.DateTimeField(blank=True, null=True, verbose_name='messaged at')),
                 ('sender_names', models.JSONField(blank=True, null=True, verbose_name='sender names')),
+                ('summary', models.TextField(blank=True, default='', verbose_name='summary')),
```
sylvinus: We can't change migrations, it must be a new file created with `make migrations`

src/backend/core/models.py
```diff
@@ -524,6 +524,7 @@ class Thread(BaseModel):
     has_active = models.BooleanField(_("has active"), default=False)
     messaged_at = models.DateTimeField(_("messaged at"), null=True, blank=True)
     sender_names = models.JSONField(_("sender names"), null=True, blank=True)
+    summary = models.TextField(_("summary"), blank=True, default="")
```
sylvinus: Default should be None / null=True ?

src/backend/core/ai/utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+def is_ai_summary_enabled() -> bool:
+    """
+    Check if AI summary features are enabled.
+    This is determined by the presence of the AI settings and if AI_FEATURE_SUMMARY_ENABLED is set to 1.
+    """
+    return (
+        is_ai_enabled()
+        and getattr(settings, "AI_FEATURE_SUMMARY_ENABLED", False) is True
+    )
```
jbpenrath: Ditto 


```suggestion
    return all([
        is_ai_enabled(),
        getattr(settings, "AI_FEATURE_SUMMARY_ENABLED", False)
    ])
```

src/backend/core/ai/utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+## Check if AI features are enabled based on settings
+
+
+def is_ai_enabled() -> bool:
+    """Check if AI features are enabled based on the presence of required settings in the environment (API_KEY, BASE_URL, MODEL)"""
+    ai_key = getattr(settings, "AI_API_KEY", None)
+    ai_base_url = getattr(settings, "AI_BASE_URL", None)
+    ai_model = getattr(settings, "AI_MODEL", None)
+    return ai_base_url is not None and ai_model is not None and ai_key is not None
```
jbpenrath: Just for the conciseness.
```suggestion
    return all([
        getattr(settings, "AI_API_KEY", None),
        getattr(settings, "AI_BASE_URL", None),
        getattr(settings, "AI_MODEL", None)
    ])
```

src/backend/core/api/viewsets/config.py
```diff
@@ -74,4 +76,7 @@ def get(self, request):
             if hasattr(settings, setting):
                 dict_settings[setting] = getattr(settings, setting)
 
+        dict_settings["AI_ENABLED"] = is_ai_enabled()
+        dict_settings["AI_FEATURE_SUMMARY_ENABLED"] = is_ai_summary_enabled()
+
```
jbpenrath, line 88: You should update the `@extend_schema` decorator to add those two new properties otherwise they will not be declared in the openapi schema so in the typescript type.

src/frontend/src/features/utils/ai-config.ts
```diff
@@ -0,0 +1,26 @@
+
+import { useConfigRetrieve } from "@/features/api/gen/config/config";
+import { ConfigRetrieve200 } from "@/features/api/gen/models/config_retrieve200";
+
+type ConfigWithAI = ConfigRetrieve200 & {
+    AI_ENABLED?: boolean;
+    AI_FEATURE_SUMMARY_ENABLED?: boolean;
+};
```
jbpenrath: This comment should prevent to have to do that
https://github.com/suitenumerique/messages/pull/286/files#r2226212365

src/frontend/src/features/utils/ai-config.tsx
jbpenrath, line 1: This file must deleted. It is duplicated with `ai-config.ts`

src/frontend/src/features/utils/mail-helper.tsx
```diff
@@ -79,6 +80,28 @@ class MailHelper {
...
+    }
+
+    /**
+     * Check if any attachment keyword is mentioned in the draft text.
+     */
+    static areAttachmentsMentionedInDraft(draftText: string): boolean {
+        const keyWordsAttachments = MailHelper.getAttachmentKeywords(DetectionMap);
+        const messageEditorDraft = draftText?.toLowerCase() || "";
+        return keyWordsAttachments.some((keyword) => messageEditorDraft.includes(keyword));
+    }
```
jbpenrath, line 104: Can you write some test for those two methods please.

src/frontend/src/features/layouts/components/thread-view/components/thread-summary/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+            ) : (
+              <p>{t("summary.no_summary")}</p>
+            )}
+          </div>
+          <div className="thread-summary__refresh-button">
+            <Tooltip content={t("actions.refresh_summary")}>
+              <Button
+                color="tertiary-text"
+                size="small"
+                icon={<span className="material-icons">autorenew</span>}
```
jbpenrath: An `Icon` component is now available in `@gouvfr/ui-kit` package you should use it.
```suggestion
                icon={<Icon name="autorenew" />}
```

PR #294, opened 2025-07-25
✨️(ai) add deep search

thomasgegout:
## Purpose

Add AI-powered intelligent email search functionality in the search bar using the Albert API from Etalab to enable semantic search capabilities beyond traditional keyword matching.

## Proposal

This PR introduces a comprehensive deep search system that allows users to find relevant emails using natural language queries, using two different methods.

### Key Features:

- **RAG Search**: use of the RAG of Albert API for deep search
- **Smart Email Processing**: Full content indexing including attachments' names and metadata
- **Contextual Search**: other method for deep search using only the context window of albert-large (mistral small)
- **Frontend Integration**: a new field "Demander à l'IA de vous aider à rechercher..." in the search bar

### DIFF
--- a/src/backend/deep_search/config.py
+++ b/src/backend/deep_search/config.py
@@ -1,0 +1,26 @@
+"""
+Configuration settings for the Albert chatbot.
+
+This module contains configuration classes and settings for the chatbot system.
+"""
+
+import os
+from dataclasses import dataclass
+from enum import Enum
+
+@dataclass
+class AlbertConfig:
+    """Configuration for Albert API."""
+    name: str = "albert-etalab"
+    base_url: str = os.getenv("AI_BASE_URL", "https://albert.api.etalab.gouv.fr/v1")
+    api_key: str = os.getenv("AI_API_KEY", "")
+    model: str = os.getenv("AI_MODEL", "albert-large")
+    temperature: float = 0.3
+    max_tokens: int = 4000  # Increased for better response capacity
+    timeout: int = 120
+    max_iterations: int = 5
+    # RAG-specific settings
+    max_emails_per_batch: int = 50  # Legacy limit (not currently used - all emails are processed)
+    max_email_content_length: int = 15000  # Limit email content size (increased to preserve attachments)
+    batch_upload_delay: float = 2.0  # Delay between batch uploads
+    min_relevance_score_percentage: float = 0.95  # Minimum score as percentage of highest score (0.0-1.0)
--- a/src/backend/deep_search/rag.py
+++ b/src/backend/deep_search/rag.py
@@ -1,0 +1,586 @@
+import os
+import json
+import requests
+import logging
+import wget
+import uuid
+import hashlib
+import time
+import tempfile
+
+from .config import AlbertConfig
+from .api_client import AlbertAPIClient
+
+logger = logging.getLogger(__name__)
+
+class RAGSystem:
+    """System for Retrieval-Augmented Generation with Albert API."""
+
+    def __init__(self):
+        """Initialize the RAG system."""
+        self.config = AlbertConfig()
+        
+        # Validate that base_url is properly configured
+        if not self.config.base_url:
+            raise ValueError("AI_BASE_URL environment variable is not set or is empty. Please configure it with the Albert API base URL.")
+        
+        if not self.config.base_url.startswith(('http://', 'https://')):
+            raise ValueError(f"Invalid base URL '{self.config.base_url}'. URL must start with http:// or https://")
+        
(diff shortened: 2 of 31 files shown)


src/backend/deep_search/config.py
```diff
@@ -0,0 +1,26 @@
...
+from dataclasses import dataclass
+from enum import Enum
+
+@dataclass
+class AlbertConfig:
+    """Configuration for Albert API."""
+    name: str = "albert-etalab"
+    base_url: str = os.getenv("AI_BASE_URL", "https://albert.api.etalab.gouv.fr/v1")
+    api_key: str = os.getenv("AI_API_KEY", "")
+    model: str = os.getenv("AI_MODEL", "albert-large")
```
jbpenrath, line 17: you should retrieve that from django settings

src/backend/deep_search/rag.py
```diff
@@ -0,0 +1,586 @@
...
+            logger.info("No previous indexing hash found, indexing needed")
+            return True
+            
+        # Reindex if email set has changed significantly
+        if current_hash != self.last_indexed_hash:
+            logger.info("Email set has changed, reindexing needed")
+            return True
+            
+        # Reindex if it's been more than 2 hours since last indexing (to handle any API issues)
+        if self.last_index_time and (current_time - self.last_index_time) > 7200:
```
jbpenrath, line 230: You could declare the duration as a class attribute
```suggestion
        if self.last_index_time and (current_time - self.last_index_time) > self.INDEX_STALE_TIME:
```

src/backend/deep_search/rag.py
```diff
@@ -0,0 +1,586 @@
...
+            if not self.embeddings_model:
+                # Fallback to a default embedding model if none found
+                self.embeddings_model = "embeddings-small"
+                
+        except Exception as e:
+            logger.error(f"Failed to get models from Albert API: {e}")
+            # Use default embedding model as fallback
+            self.embeddings_model = "embeddings-small"
+
+    def collection_exists(self, user_id: str) -> bool:
```
jbpenrath, line 121: this method is duplicated

src/backend/deep_search/rag.py
```diff
@@ -0,0 +1,586 @@
...
+        try:
+            logger.info(f"Adding single email {email_id} to collection {self.collection_id}")
+            
+            # Prepare document for indexing
+            documents = [{
+                "id": email_id,
+                "content": email.get('body', ''),
+                "metadata": {
+                    "subject": email.get('subject', ''),
+                    "sender": email.get('sender', ''),
```
jbpenrath, line 363: It could interesting to index recipients nope ? (to and cc)
Maybe also if email has attachments just through a boolean value ?

src/backend/deep_search/views.py
```diff
@@ -0,0 +1,283 @@
...
+                'success': False,
+                'error': 'query is required'
+            }, status=status.HTTP_400_BAD_REQUEST)
+        
+        # Get user ID from authenticated user
+        if not hasattr(request, 'user') or not request.user.is_authenticated:
+            return Response({
+                'success': False,
+                'error': 'Authentication required'
+            }, status=status.HTTP_401_UNAUTHORIZED)
```
jbpenrath, line 60: If user is not authenticated the permission `IsAuthenticated` should have already return a 401 response.

src/backend/deep_search/views.py
```diff
@@ -0,0 +1,283 @@
...
+            
+            # Get chatbot instance for API client
+            chatbot = get_chatbot()
+            
+            # Perform intelligent email search using RAG
+            search_results = email_service.chatbot_intelligent_email_search(
+                user_id=user_id,
+                user_query=query,
+                api_client=chatbot.api_client,
+                max_results=10
```
jbpenrath, line 78: Could we imagine to let api consumer to set this property through a query param? Or is it a threshold for performance purpose?

src/backend/deep_search/views.py
```diff
@@ -0,0 +1,283 @@
...
+from rest_framework.response import Response
+from rest_framework import status
+
+from .chatbot import get_chatbot
+from .email_service import EmailService
+
+logger = logging.getLogger(__name__)
+
+
+@api_view(['POST'])
```
jbpenrath, line 21: IMO this should be a get endpoint not a post as it does not create resource then the query can be easily get from a query param

src/backend/deep_search/views.py
jbpenrath, line 1: You tried to standardize response format but currently it will really hard to maintain. IMO a DRF Serializer could help to factorize that.

PR #303, opened 2025-08-04
✨(templates) add message templates for signatures and replies

sdemagny:
- [x] schema openapi
- [x] FIXME filter user accesses
- [x] MessageTemplateKindChoices
- [x] Fix potential security issues with signature replacement

## Summary by CodeRabbit

* **New Features**
  * Global message templates ("signatures") — create, render, auto-apply and attach to drafts and outgoing messages; editor support to insert/manage signatures and template variables.

* **API**
  * New template endpoints and schemas; draft/send payloads accept signatureId; messages now include signature info; mailbox access checks enforced.

* **UI**
  * Admin "Signatures" area (tab, list, compose modal, datagrid); composer/editor signature selector; composer translations and toasts integrated.

* **Tests**
  * Extensive coverage for templates, rendering, draft and send signature flows.

* **Migrations**
  * DB migration adding message templates, blob ownership, and per-message signature field.


src/backend/core/models.py
```diff
@@ -1345,3 +1347,163 @@ def get_dns_record_value(self) -> str:
...
+            "Type of template (reply, forward, new message, auto reply, signature)"
+        ),
+    )
+
+    content_type = models.SmallIntegerField(
+        _("content type"),
+        choices=SignatureTypeChoices.choices,
+        default=SignatureTypeChoices.TEXT,
+        help_text=_("Type of content (text, HTML, rich text) - mainly for signatures"),
+    )
```
sdemagny, line 1399: delete

src/backend/core/models.py
```diff
@@ -1345,3 +1347,163 @@ def get_dns_record_value(self) -> str:
...
+
+    subject = models.CharField(
+        _("subject"),
+        max_length=255,
+        blank=True,
+        help_text=_(
+            "Subject template with placeholders (e.g., 'Re: {original_subject}') - not used for signatures"
+        ),
+    )
```
sdemagny, line 1376: delete

src/backend/core/models.py
```diff
@@ -1345,3 +1347,163 @@ def get_dns_record_value(self) -> str:
...
+        blank=True,
+        related_name="created_templates",
+        help_text=_("User who created this template"),
+    )
+
+    class Meta:
+        db_table = "messages_emailtemplate"
+        verbose_name = _("email template")
+        verbose_name_plural = _("email templates")
+        unique_together = ("name", "mailbox")
```
sdemagny, line 1440: (name, maildomain)

src/backend/core/models.py
```diff
@@ -1345,3 +1346,195 @@ def get_dns_record_value(self) -> str:
         """Get the DNS TXT record value for this DKIM key."""
         algorithm_enum = DKIMAlgorithmChoices(self.algorithm)
         return f"v=DKIM1; k={algorithm_enum.label}; p={self.public_key}"
+
+
+class EmailTemplate(BaseModel):
```
sylvinus: Why "EmailTemplate" and not "MessageTemplate" ? Or even just "Template"

src/backend/core/models.py
```diff
@@ -1345,3 +1346,195 @@ def get_dns_record_value(self) -> str:
...
+            return self.html_body or self.text_body
+
+        # For non-signatures, return HTML body if available, otherwise text body
+        return self.html_body or self.text_body
+
+    def is_signature(self) -> bool:
+        """Check if this template is a signature."""
+        return self.type == TemplateTypeChoices.SIGNATURE
+
+    def is_template(self) -> bool:
```
sylvinus: Template.is_template also feels sub-optimal naming

src/backend/core/models.py
sdemagny: MessageTemplate 👍

src/backend/core/admin.py
```diff
@@ -54,6 +54,22 @@ def reset_keycloak_password_action(_, request, queryset):
 )
 
 
+def duplicate_template_action(_, request, queryset):
```
sylvinus: Do we need this in the admin?

src/backend/core/enums.py
```diff
@@ -77,3 +77,13 @@ class UserAbilityChoices(models.TextChoices):
 
     CAN_VIEW_DOMAIN_ADMIN = "view_maildomains", "Can view domain admin"
     CAN_CREATE_MAILDOMAINS = "create_maildomains", "Can create maildomains"
+
+
+class TemplateTypeChoices(models.IntegerChoices):
+    """Defines the possible types of email templates."""
+
+    REPLY = 1, "reply"
+    FORWARD = 2, "forward"
```
sylvinus: Forward?

src/backend/core/models.py
sylvinus: Also it seems to never be called. And I'm not sure just for the admin we need a is_signature() function

src/backend/core/models.py
```diff
@@ -1345,3 +1346,195 @@ def get_dns_record_value(self) -> str:
...
+        Mainly used for signatures.
+
+        Returns:
+            Formatted template content
+        """
+        if self.type == TemplateTypeChoices.SIGNATURE:
+            # For signatures, prefer HTML body if available, otherwise text body
+            return self.html_body or self.text_body
+
+        # For non-signatures, return HTML body if available, otherwise text body
```
sylvinus: There is no difference?

src/backend/core/enums.py
```diff
@@ -77,3 +77,13 @@ class UserAbilityChoices(models.TextChoices):
...
+    NEW_MESSAGE = 3, "new_message"
```
sylvinus: If we have templates for new messages, we should also need a Template.subject field ?

src/backend/core/admin.py
sdemagny: It's a super handy action for testing things myself in the Django admin :D I have a feeling we're going to need it to initialize some data.

src/backend/core/enums.py
sdemagny: We discussed it with JB. For now, there will be a template selection at the text block level, so I removed the subject field. Honestly, I'm a bit out of ideas for what this field could be used for. I'm open to suggestions.

src/backend/core/models.py
```diff
@@ -1389,3 +1390,182 @@ def get_dns_record_value(self) -> str:
         """Get the DNS TXT record value for this DKIM key."""
         algorithm_enum = DKIMAlgorithmChoices(self.algorithm)
         return f"v=DKIM1; k={algorithm_enum.label}; p={self.public_key}"
+
+
+class MessageTemplate(BaseModel):
```
jbpenrath, line 1641: IMO we should also store the raw block note object like we do for draft message. WDYT to add a field `raw_body` or `raw_blob` ?

src/backend/core/api/serializers.py
```diff
@@ -887,3 +887,112 @@ class ImportIMAPSerializer(ImportBaseSerializer):
...
+                )
+
+                # Exclude current instance if updating
+                if self.instance:
+                    existing_default = existing_default.exclude(pk=self.instance.pk)
+
+                if existing_default.exists():
+                    raise serializers.ValidationError(
+                        f"Another default {template_kind} template already exists."
+                    )
```
jbpenrath: According to comments, I understand you mixed up active flag with default flag nope?

We could and must have several active template.
And about the default flag, IMO the first template of a given `kind` should be set as default then other should not be considered to default. Then if we user create a new template and ask to be the default of its kind, the current default should be demoted.

src/backend/core/models.py
sylvinus: You mean we should store both the raw (for being able to edit) and the rendered (to be able to send/assemble from the backend) ?

src/backend/core/models.py
```diff
@@ -1389,3 +1390,182 @@ def get_dns_record_value(self) -> str:
...
+
+    text_body = models.TextField(
+        _("text body"),
+        blank=True,
+        help_text=_(
+            "Plain text email body template with placeholders (e.g., 'Dear {recipient_name},') or signature content"
+        ),
+    )
+
+    kind = models.SmallIntegerField(
```
sylvinus: I think we use "type" elsewhere

src/backend/core/models.py
jbpenrath: Actually, I wonder if we do not just store the `raw_body` from the backend. Then about the rendering, as we need to render it in the message composer in the frontend, I think we could render the signature html and text string from the frontend WDYT ?

jbpenrath: Actually, I wonder if we do not just store the raw_body from the backend. Then about the rendering, as we need to render it in the message composer in the frontend, I think we could embedded the signature in the message body from the frontend when we convert the blocknote content into html and text string WDYT ?

sylvinus: I have 2 issues with that:
 - The user could potentially inject unwanted signatures
 - Beyond signatures, if we have auto-replies for instance, we won't always have a frontend in the loop

src/backend/core/api/viewsets/message_template.py
```diff
@@ -0,0 +1,185 @@
...
+    - is_default: Filter by default status (true/false) - works with mailbox or maildomain
+    """
+
+    permission_classes = [permissions.IsAuthenticated]
+    serializer_class = MessageTemplateSerializer
+    pagination_class = None  # Disable pagination
+    filterset_fields = [
+        "kind",
+        "is_active",
+    ]
```
jbpenrath: This class attribute seems useless.

src/backend/core/models.py
```diff
@@ -1401,3 +1402,191 @@ def get_dns_record_value(self) -> str:
...
+        help_text=_(
+            "Name of the template (e.g., 'Standard Reply', 'Out of Office', 'Work Signature')"
+        ),
+    )
+
+    description = models.TextField(
+        _("description"),
+        blank=True,
+        help_text=_("Optional description of when to use this template"),
+    )
```
jbpenrath: What do you think to make this field optional? (`default=""` ?)

src/backend/core/api/serializers.py
jbpenrath, line 1: On MessageTemplate creation and update, you should manage the `rawBlob` field like we did for `Message`.
https://github.com/suitenumerique/st-messages/blob/0c79a169aaa0fb47c4f7e14871dffa033e493e62/src/backend/core/api/viewsets/draft.py#L427-L430
https://github.com/suitenumerique/st-messages/blob/0c79a169aaa0fb47c4f7e14871dffa033e493e62/src/backend/core/api/viewsets/draft.py#L254-L265

Currently, this field expect a blob id as value but I would like to provide the value store into the blob.

src/backend/core/models.py
```diff
@@ -1401,3 +1402,191 @@ def get_dns_record_value(self) -> str:
...
+
+    raw_blob = models.ForeignKey(
+        "Blob",
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+        related_name="message_templates",
+        help_text=_("Reference to the blob containing raw template data"),
+    )
```
jbpenrath: Currently, it will be hard to create a blob for MessageTemplate because the `Blob` model must be linked to a `Mailbox`.

src/backend/core/api/viewsets/message_template.py
```diff
@@ -0,0 +1,185 @@
...
+                )
+        if maildomain_id:
+            queryset = queryset.filter(
+                template_maildomains__maildomain_id=maildomain_id
+            )
+            if is_default is not None:
+                is_default_bool = is_default.lower() in ("true", "1", "yes")
+                queryset = queryset.filter(
+                    template_maildomains__is_default=is_default_bool
+                )
```
jbpenrath: When the user specifies a `mailbox_id`, `MessageTemplate` linked to the maildomain's mailbox should be also returned.

src/backend/core/api/viewsets/message_template.py
```diff
@@ -0,0 +1,185 @@
...
+        except (Mailbox.DoesNotExist, ValueError):
+            return Response(
+                {"error": "Failed to render template: Invalid mailbox ID"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        context = {
+            "mailbox": str(mailbox),
+            "fullname": user.full_name,
+        }
```
jbpenrath: This should be computed according to available placeholders.

src/backend/core/models.py
sdemagny: description is already optional. `blank=True`

src/backend/core/api/viewsets/message_template.py
sdemagny: I forgot to modify this part after implementing the custom attributes, sorry

src/backend/core/models.py
jbpenrath: Yup but as there is no default value, this is mandatory to provide a value.

src/backend/core/models.py
```diff
@@ -1401,3 +1460,184 @@
...
+                fields=("maildomain", "type"),
+                condition=models.Q(is_forced=True),
+                name="uniq_forced_template_maildomain_type",
+            ),
+        ]
+        indexes = [
+            models.Index(fields=("mailbox", "type", "is_active")),
+            models.Index(fields=("maildomain", "type", "is_active")),
+        ]
+
```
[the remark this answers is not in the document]
sdemagny: reject

src/frontend/src/features/forms/components/message-composer/index.tsx
```diff
@@ -59,19 +65,20 @@ export const MessageComposer = ({ mailboxId, blockNoteOptions, defaultValue, quo
...
+        // Parse initial content
         const initialContent = defaultValue
             ? JSON.parse(defaultValue)
             : [{ type: "paragraph", content: "" }];
 
-        if (!quotedMessage) return initialContent;
-
-        return initialContent.concat([{
+        // Add quoted message if needed
+        const contentWithQuote = quotedMessage ? initialContent.concat([{
```
jbpenrath: IMO this is more confusing to read.

src/frontend/src/features/forms/components/message-composer/index.tsx
```diff
@@ -101,49 +111,113 @@ export const MessageComposer = ({ mailboxId, blockNoteOptions, defaultValue, quo
         ...blockNoteOptions,
     }, [i18n.resolvedLanguage]);
 
-    const handleChange = async () => {
-        const markdown = await editor.blocksToMarkdownLossy(editor.document);
+    const handleChange = useCallback(async () => {
+        // Filter out signature blocks from content before converting to markdown/HTML
+        const contentWithoutSignature = editor.document.filter(block => block.type !== 'signature');
```
jbpenrath: This is not the best way to exclude signature block from rendering.

Instead you should the same thing we did for th `quoted-messsage-block` : Override the `toExternalHTML` method as following :
```
toExternalHTML: () => (<span />)
```

src/frontend/src/features/blocknote/signature-block/index.tsx
```diff
@@ -10,18 +11,39 @@ type SignatureTemplateSelectorProps = {
...
+        
+        // Update the form value - use null instead of undefined for explicit removal
+        const valueToSet = currentSignatureId === undefined ? null : currentSignatureId;
+        form.setValue('signatureId', valueToSet as string | null, { shouldDirty: true });
+        
+        // Trigger the MessageComposer's handleChange which will save the draft
+        if (onSignatureChange) {
+            onSignatureChange();
+        }
+    };
```
jbpenrath: With those changes, this block is IMO too muck link to the `MessageComposer` component and it introduces side effect that is weird.
IMO this component should not be responsible to update the formValue. All those things should be done through the callback.

src/frontend/src/features/blocknote/signature-block/index.tsx
```diff
@@ -60,24 +82,53 @@ export const SignatureTemplateSelector = ({ mailboxId, templates = [], isLoading
...
                 } else {
+                    // Insert signature at the end of the document
+                    if (editor.document.length === 0) {
+                        // If document is empty, first add an empty paragraph
+                        editor.insertBlocks(
+                            [{ type: "paragraph", content: [{ type: "text", text: "", styles: {} }] }] as unknown as PartialMessageComposerBlockSchema[],
+                            "",
+                            "after"
+                        );
+                    }
```
jbpenrath, line 124: IMO this is not the responsibility of this component to do that. This should be the responsibility of the editor.

src/frontend/src/features/forms/components/message-composer/index.tsx
```diff
@@ -101,49 +111,113 @@ export const MessageComposer = ({ mailboxId, blockNoteOptions, defaultValue, quo
...
-                    ["signature"],
-                    [{ id: "signature", type: "signature", props: { templateId: defaultSignatureTemplateId, mailboxId: mailboxId } }],
-                );
+        if (!editor || isLoadingSignatures || activeSignatures.length === 0) return;
+
+        // Check if signature is already in the editor
+        const existingSignature = editor.document.find(block => block.type === 'signature');
+        if (existingSignature) {
+            return;
+        }
```
jbpenrath: Why don't use `editor.getBlock('signature');` ? The `id` of a block is set programmatically so we should be confident that a block with `id` equals to `signature` is of type `signature`

src/frontend/src/features/forms/components/message-composer/index.tsx
```diff
@@ -101,49 +111,113 @@ export const MessageComposer = ({ mailboxId, blockNoteOptions, defaultValue, quo
...
+        if (draft?.signature?.id && activeSignatures.some(sig => sig.id === draft.signature?.id)) {
+            signatureToUse = draft.signature;
+        } 
+        // Priority 2: Forced signature (if no draft signature)
+        else {
+            const forcedSignature = activeSignatures.find(signature => signature.is_forced);
+            if (forcedSignature) {
+                signatureToUse = forcedSignature;
             }
         }
```
jbpenrath:
```suggestion
        let signatureToUse = undefined;

        // Priority 1: Draft signature (if exists and is still active)
        if (draft?.signature?.id && activeSignatures.some(sig => sig.id === draft.signature?.id)) {
            signatureToUse = draft.signature;
        } 
        // Priority 2: Forced signature (if no draft signature)
        else {
            signatureToUse = activeSignatures.find(signature => signature.is_forced);
        }
```

src/frontend/src/features/forms/components/message-composer/index.tsx
```diff
@@ -101,49 +111,113 @@ export const MessageComposer = ({ mailboxId, blockNoteOptions, defaultValue, quo
...
-        // In case there is a signature block but the templateId does not match an active signature, we remove the block.
-        if (signatureBlock) {
-            const isSignatureActive = activeSignatures.findIndex(signature => signature.id === (signatureBlock.props as BlockSignatureConfigProps).templateId);
-            if (isSignatureActive === -1) editor.removeBlocks(["signature"]);
```
jbpenrath: You removed this case but I don't see where it is managed now

src/frontend/src/features/forms/components/message-form/index.tsx
```diff
@@ -381,16 +413,24 @@ export const MessageForm = ({
             return;
         }
 
-        const draft = await saveDraft(data);
+        // Ensure any pending auto-save operations complete before sending
+        // This prevents race conditions where editor changes haven't been saved yet
+        await new Promise(resolve => setTimeout(resolve, 350)); // Wait for debounce + buffer
```
jbpenrath: This is not bullet proof. Instead if `isSaving` is `true` we could just recall `handleSubmit` through a timeout WDYT ?

sdemagny:
```suggestion
        if (isSaving) {
            // If currently saving, wait and retry submit
            setTimeout(() => handleSubmit(data), 500);
            return;
        }
```
?

src/frontend/src/features/forms/components/message-composer/index.tsx
```diff
@@ -110,9 +110,9 @@ export const MessageComposer = ({ mailboxId, blockNoteOptions, defaultValue, quo
 
     const handleChange = useCallback(async () => {
         // Filter out signature blocks from content before converting to markdown/HTML
-        const contentWithoutSignature = editor.document.filter(block => block.type !== 'signature');
+        //const contentWithoutSignature = editor.document.filter(block => block.type !== 'signature');
```
jbpenrath: Just remove that
```suggestion
```

src/frontend/src/features/forms/components/message-composer/index.tsx
```diff
@@ -237,7 +237,52 @@ export const MessageComposer = ({ mailboxId, blockNoteOptions, defaultValue, quo
...
+                                    );
+                                }
+                            }
+
+                            // Update form value and trigger draft save
+                            form.setValue('signatureId', signatureId, { shouldDirty: true });
+                            if (onSaveDraft) {
+                                onSaveDraft();
+                            }
+                        }}
```
jbpenrath: Sorry my previous comment was maybe unclear but IMO all the logic in charge to add/remove signature block must stay into `signature-block` but the logic in charge to update form state and persist draft on signature change should be the responsibility to the `MessageComposer`.

src/backend/core/admin.py
```diff
@@ -610,3 +610,34 @@ class DKIMKeyAdmin(admin.ModelAdmin):
             },
         ),
     )
+
+
+@admin.register(models.MessageTemplate)
+class MessageTemplateAdmin(admin.ModelAdmin):
```
sylvinus, line 643: - [x] we should have `autocomplete_fields` for maildomain & mailbox
- [x] same for MessageAdmin.signature
- [x] we have preview_content on the list but not on the detail ? Shouldn't it be the reverse ?

src/backend/core/api/serializers.py
```diff
@@ -299,6 +303,62 @@ def get_name(self, instance):
         return None
 
 
+class ReadOnlyMessageTemplateSerializer(serializers.ModelSerializer):
+    """Serialize message templates for read-only operations."""
+
+    type = IntegerChoicesField(choices_class=models.MessageTemplateTypeChoices)
```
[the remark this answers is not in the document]
sylvinus: @sdemagny

src/backend/core/api/serializers.py
```diff
@@ -299,6 +303,62 @@ def get_name(self, instance):
...
+    html_body = serializers.SerializerMethodField()
+    text_body = serializers.SerializerMethodField()
+    raw_body = serializers.SerializerMethodField()
+
+    def get_html_body(self, obj) -> str:
```
sylvinus, line 314: This should use models.MessageTemplate.html_body ?

src/backend/core/models.py
```diff
@@ -1443,3 +1511,205 @@ def get_dns_record_value(self) -> str:
...
+                placeholder, escape(str(value))
+            )
+            rendered_text_body = rendered_text_body.replace(placeholder, str(value))
+
+        return {
+            "html_body": rendered_html_body,
+            "text_body": rendered_text_body,
+        }
+
+    def get_formatted_content(self) -> str:
```
sylvinus: not a big fan of this method, if we want to do "html or text", which is slightly questionable, it should only be done in the admin method

src/backend/core/api/viewsets/message_template.py
```diff
@@ -0,0 +1,232 @@
...
+        description="Render a template with the provided context variables.",
+    )
+    @action(detail=True, methods=["get"], url_path="render")
+    def render_template(self, request, pk=None):  # pylint: disable=unused-argument
+        """Render a template."""
+
+        # get template
+        template = self.get_object()
+        try:
+            rendered = template.render_template(request.user)
```
sylvinus: as discussed together, render should not take the user as argument

src/backend/core/api/viewsets/message_template.py
```diff
@@ -0,0 +1,232 @@
...
+from core.api import permissions
+from core.api.serializers import (
+    MessageTemplateSerializer,
+    ReadOnlyMessageTemplateSerializer,
+)
+from core.models import Mailbox, MailDomain, MessageTemplate, MessageTemplateTypeChoices
+
+
+@extend_schema(tags=["message-templates"])
+class MessageTemplateViewSet(viewsets.ModelViewSet):  # pylint: disable=too-many-ancestors
```
sylvinus: as discussed together, the viewset seems to re-implement access control in a addition to other places, try to factor it in a single place (model?)

src/backend/core/mda/inbound.py
```diff
@@ -12,7 +12,7 @@
 from django.db.utils import Error as DjangoDbError
 from django.utils import timezone
 
-from core import models
+from core import enums, models
```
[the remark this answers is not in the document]
sylvinus: @sdemagny

src/backend/core/api/serializers.py
```diff
@@ -1095,3 +1158,185 @@ class ImportIMAPSerializer(ImportBaseSerializer):
...
+            if any(field is not None for field in [html_body, text_body, raw_body]):
+                # Get current content
+                current_content = (
+                    instance.blob.get_content().decode("utf-8")
+                    if instance.blob
+                    else {"html": "", "text": "", "raw": None}
+                )
+
+                # Update with new values, keeping existing values if not provided
+                # Parse raw_body if it's a JSON string
```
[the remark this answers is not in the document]
sylvinus: @sdemagny indeed ? aren't we missing a json.loads ? and also shouldn't we use instance.html_body directly?

src/backend/core/api/serializers.py
```diff
@@ -1095,3 +1158,186 @@ class ImportIMAPSerializer(ImportBaseSerializer):
...
+                    "html": html_body
+                    if html_body is not None
+                    else current_content["html"],
+                    "text": text_body
+                    if text_body is not None
+                    else current_content["text"],
+                    "raw": raw_body,
+                }
+
+                # Delete old blob
```
sylvinus: Here, as we indeed delete the old blob on updates, I don't understand why we need to read its content before ? Each update should have the 3 fields and overwrite the full content?

src/backend/core/mda/draft.py
```diff
@@ -22,6 +106,8 @@ def create_draft(
     cc_emails: Optional[list] = None,
     bcc_emails: Optional[list] = None,
     attachments: Optional[list] = None,
+    signature_id: Optional[str] = None,
+    user: Optional[models.User] = None,
```
sylvinus: will we also remove this user argument?

sylvinus (changes requested):
Backend feedbacks

src/frontend/src/features/utils/mail-helper.tsx
```diff
@@ -41,7 +41,9 @@ class MailHelper {
      * then render HTML ready for email through react-email.
      */
     static async markdownToHtml(markdown: string) {
-        return renderToString(<Markdown>{markdown}</Markdown>);
+        return renderToString(<Markdown>{markdown}</Markdown>)
+            .replace(/(^<div data-id="react-email-markdown">|<\/div>$)/g, '')
```
sylvinus, line 45: if we don't have `<div data-id="react-email-markdown">` anymore, this could remove a closing `</div>`

src/backend/core/mda/draft.py
sdemagny: It's necessary to perform a final user access check in `mailbox.get_validated_signature`

src/backend/core/models.py
```diff
@@ -665,6 +631,87 @@ def get_abilities(self, user):
...
+            if not in_sender_scope:
+                logger.warning(
+                    "User %s attempted to use signature %s outside sender scope",
+                    user.id,
+                    signature_id,
+                )
+                return None
+
+            # Verify user has access to the signature template
+            user_has_access = False
```
sylvinus: @sdemagny I still don't understand why we need to pass `user` and check this here? If we remove all these checks the tests still pass.

src/backend/core/mda/outbound.py
```diff
@@ -64,6 +65,37 @@ def prepare_outbound_message(
...
+                mailbox=mailbox_sender, user=user
+            )
+            if signatures:
+                text_body = (
+                    text_body + "\n" + signatures["text_body"]
+                    if text_body
+                    else signatures["text_body"]
+                )
+                html_body = (
+                    html_body + "<p>" + signatures["html_body"] + "</p>"
```
sylvinus, line 87: Do we need to wrap in a `<p>` ? Won't this produce invalid markup with some tags in html_body ?

PR #304, opened 2025-08-05
♻️(services) refactor Services into a single directory

sylvinus:
I haven't done AI yet because there are pending patches but once done we'll also have `core/services/ai/*.py`

I used `importer` as a name to avoid using a Python keyword


sdemagny (approved):
\o/

PR #319, opened 2025-08-19
Selfcheck prometheus

Nastaliss:
## Purpose

Add prometheus metrics that get pushed to a [Prometheus pushgateway](https://github.com/prometheus/pushgateway) on success.

## Proposal
By default, the script uses the `SelfCheckMetricsBase` class as its metrics collector, that does not do anything on start, stop, etc. 
If the Prometheus env variables are defined, the metrics collector will use `SelfCheckPrometheusMetrics` that extends `SelfCheckMetricsBase` and will update its metrics and push them to the pushgateway on exit. 

This allows users to extend the MetricsBase class to integrate their own system, as well as having a single `if` clause that does not render the rest of the code unreadable.

### DIFF
--- a/src/backend/core/management/commands/selfcheck.py
+++ b/src/backend/core/management/commands/selfcheck.py
@@ -34,18 +34,18 @@ class SelfCheckMetricsBase(object):
 
 
 class SelfCheckPrometheusMetrics(SelfCheckMetricsBase):
-    """Prometheus metrics for the selfs-check process."""
+    """Prometheus metrics for the selfcheck process."""
 
     def __init__(self):
         prefix = settings.MESSAGES_SELFCHECK_PROMETHEUS_METRICS_PREFIX
         if settings.MESSAGES_SELFCHECK_PROMETHEUS_METRICS_PUSHGATEWAY_URL is None:
             raise ValueError("Prometheus push gateway URL is not set.")
         self.registry = CollectorRegistry()
-        self.start_time = Gauge(f'{prefix}self_check_start_time', 'Start timestamp of the self check', registry=self.registry)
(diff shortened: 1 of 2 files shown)


src/backend/core/management/commands/selfcheck.py
```diff
@@ -4,9 +4,70 @@
...
+
+    def write_reception_time(self, reception_time):
+        pass
+
+    def send_metrics(self):
+        pass
+
+
+class SelfCheckPrometheusMetrics(SelfCheckMetricsBase):
+    """Prometheus metrics for the selfs-check process."""
```
sylvinus: selfcheck

src/backend/core/management/commands/selfcheck.py
```diff
@@ -4,9 +4,70 @@
...
+
+    def __init__(self):
+        prefix = settings.MESSAGES_SELFCHECK_PROMETHEUS_METRICS_PREFIX
+        if settings.MESSAGES_SELFCHECK_PROMETHEUS_METRICS_PUSHGATEWAY_URL is None:
+            raise ValueError("Prometheus push gateway URL is not set.")
+        self.registry = CollectorRegistry()
+        self.start_time = Gauge(f'{prefix}self_check_start_time', 'Start timestamp of the self check', registry=self.registry)
```
sylvinus: self_check => selfcheck for greppability

sylvinus:
LGTM, I imagine we'll refactor this if we have many more prom integrations elsewhere in Messages but OK as a first step!

Nastaliss:
> LGTM, I imagine we'll refactor this if we have many more prom integrations elsewhere in Messages but OK as a first step!

Definitely, although this one is a bit special as it's pushing in a PushGateway and not exposing metrics as we'll do in the backend

PR #320, opened 2025-08-19
Feat: Add prometheus metrics

Nastaliss:
## Purpose
Add prometheus metrics using [django-prometheus](https://github.com/django-commons/django-prometheus)

## Proposal

Use the default exporter + some custom metrics


src/backend/messages/urls.py
```diff
@@ -46,3 +46,11 @@
             name="redoc-schema",
         ),
     ]
+
+if settings.ENABLE_PROMETHEUS:
```
sylvinus, line 56: should the prometheus metrics endpoint be password-protected ?

Nastaliss: Good question, we can either have a password protection or a ip-filter to only allow http calls from inside the same subnet

Nastaliss: As seen with @zeylos, we'll use an api key for now.
Fixed in [e71841693af4ff0c957aec41ba1e41ed1eb33e09](https://github.com/suitenumerique/messages/pull/320/commits/e71841693af4ff0c957aec41ba1e41ed1eb33e09)

src/backend/messages/settings.py
```diff
@@ -549,6 +549,11 @@ class Base(Configuration):
 
     ENABLE_DOCKERFLOW = values.BooleanValue(default=True)
 
+    ENABLE_PROMETHEUS = values.BooleanValue(default=True)
+    if ENABLE_PROMETHEUS:
```
sylvinus: Not tested, but I think this will always be true. The "if" block should be in an __init__() method instead.

src/backend/messages/settings.py
```diff
@@ -549,6 +549,11 @@ class Base(Configuration):
...
+        MIDDLEWARE += ["core.middlewares.BasicAuthMiddleware"]
+        PROMETHEUS_API_KEY = values.Value(None, environ_name="PROMETHEUS_API_KEY", environ_prefix=None)
```
sylvinus: this can be outside of the if block

sylvinus:
@coderabbitai review plz

src/backend/core/metrics.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        total_size = (
+            Attachment.objects.aggregate(models.Sum("blob__size"))["blob__size__sum"]
+            or 0
+        )
+        yield GaugeMetricFamily(
+            "attachments_total_size_bytes",
+            "Total size of all attachments in bytes",
+            value=total_size,
+        )
```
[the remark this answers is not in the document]
Nastaliss: Getting an error is better than getting garbage data

src/backend/core/middlewares.py
```diff
@@ -0,0 +1,18 @@
...
+    def __call__(self, request):
+        if request.path == "/metrics":
+            expected_api_key = f"Bearer {settings.PROMETHEUS_API_KEY}"
+            if (
+                settings.PROMETHEUS_API_KEY
+                and request.headers.get("Authorization") != expected_api_key
+            ):
+                return HttpResponse("Unauthorized", status=401)
+
+        return self.get_response(request)
```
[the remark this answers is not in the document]
Nastaliss: Looks like this adds a lot of complexity for a marginal gain in usability

src/backend/core/factories.py
```diff
@@ -215,3 +217,41 @@ def threads(self, create, extracted, **kwargs):
...
+    """A factory to random attachments for testing purposes."""
+
+    class Meta:
+        model = models.Attachment
+
+    mailbox = factory.SubFactory(MailboxFactory)
+    name = factory.Sequence(lambda n: f"attachment{n}.txt")
+    blob_size = 1500
+    sha256 = factory.LazyAttribute(lambda o: hashlib.sha256(o.raw_content).digest())
+
```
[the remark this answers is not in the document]
Nastaliss: https://github.com/suitenumerique/messages/pull/320#discussion_r2316697924

PR #321, opened 2025-08-19
✨(api) add /placeholders endpoint

sdemagny:
This endpoint will be used for Signature and other message templates.
Update Dockerfile to include gettext for i18n compilation


src/backend/core/tests/api/test_placeholder.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+    def test_internationalization_support(self, api_client):
+        """Test that internationalization is properly handled."""
+        url = reverse("placeholders")
+        response = api_client.get(url, HTTP_ACCEPT_LANGUAGE="fr-fr")
+        assert response.status_code == status.HTTP_200_OK
+        data = response.json()
+        assert data["job_title"] == "Fonction"
+        assert data["is_elected"] == "Est élu"
+        assert data["full_name"] == "nom complet"
```
jbpenrath: Can we capitalize those strings ?

PR #324, opened 2025-08-26
Improve development docker compose

Nastaliss:
## Purpose

When using the development stack, I noticed some minor issues in the docker compose file:
- there are no health-checks in the service to make sure it can properly start. Running `make start` would exit without error while the backend is not ready
- running `make stop` will **not** stop all services. For example, the _backend-db_ service started with make _start-minimal_ is left dangling after running make stop (and its port is still allocated).

## Proposal

- Install curl on the dev images, and add health-check that curl the base api route to make sure there are no errors preventing DRF from starting the server.
- Run `docker compose down` with the option `--profile "*"` to make sure every containers in the compose file are stopped.

## Summary by CodeRabbit

- New Features
  - Added a /healthz endpoint for quick service health verification.
  - Start commands now wait until services report healthy before returning.

- Chores
  - Added container healthchecks for DB and backend.
  - Stop command now halts all profiles for a more reliable shutdown.
  - Reduced server log noise by silencing routine health-check requests.
  - Simplified the OpenAPI schema update script’s environment.
  - Removed an unused dependency and disabled automatic DockerFlow wiring.

### DIFF
--- a/src/backend/messages/urls.py
+++ b/src/backend/messages/urls.py
@@ -34,7 +34,7 @@ if settings.DEBUG:
 if settings.USE_SWAGGER or settings.DEBUG:
     urlpatterns += [
         path(
-            f"{settings.API_VERSION}/swagger.json",
+            f"api/{settings.API_VERSION}/swagger.json",
             SpectacularJSONAPIView.as_view(
                 api_version=settings.API_VERSION,
                 urlconf="core.urls",
@@ -42,12 +42,12 @@ if settings.USE_SWAGGER or settings.DEBUG:
             name="client-api-schema",
         ),
         path(
-            f"{settings.API_VERSION}/swagger/",
+            f"api/{settings.API_VERSION}/swagger/",
             SpectacularSwaggerView.as_view(url_name="client-api-schema"),
             name="swagger-ui-schema",
         ),
         re_path(
-            f"{settings.API_VERSION}/redoc/",
+            f"api/{settings.API_VERSION}/redoc/",
             SpectacularRedocView.as_view(url_name="client-api-schema"),
             name="redoc-schema",
         ),
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@
     Roadmap
   </a> - <a href="#getting-started-">
     Getting started
-  </a> - <a href="mailto:sylvain.zimmer@beta.gouv.fr">
+  </a> - <a href="mailto:contact@suite.anct.gouv.fr">
     Reach out
   </a>
 </p>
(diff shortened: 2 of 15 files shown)


sylvinus:
Look good. Should we make sure the healthchecks won't spam the logs ?

src/backend/messages/urls.py
```diff
@@ -15,6 +16,11 @@
 urlpatterns = [
     path(settings.ADMIN_URL, admin.site.urls),
     path("", include("core.urls")),
+    path(
+        f"{settings.API_VERSION}/healthz/",
```
sylvinus: Just /healthz without the v1.0 would be better I think

src/backend/messages/wsgi.py
```diff
@@ -11,6 +11,21 @@
...
 from configurations.wsgi import get_wsgi_application
 
+# pylint: disable=all
+# Suppress access logs for healthcheck route
+try:
+    import django.core.servers.basehttp
+    class QuietWSGIRequestHandler(django.core.servers.basehttp.WSGIRequestHandler):
+        def log_message(self, format, *args):
+            path = getattr(self, "path", "")
+            if path.endswith("/healthz") or path.endswith("/healthz/"):
```
sylvinus: do a strict equality test to avoid malicious queries bypassing logs

src/backend/Dockerfile
```diff
@@ -98,6 +98,14 @@ COPY --from=base-with-deps-dev /venv /venv
 # ENV DB_HOST=postgresql \
 #     DB_PORT=5432
 
+# Switch to root to install packages
+USER root
+RUN apt-get update && apt-get install -y --no-install-recommends curl \
```
sylvinus: this could be done as part of the `base` stage, on top of the file ?

sylvinus: (discard that, it's good that curl is only in the runtime-dev stage)

sylvinus (approved):
great job

src/backend/Dockerfile
sylvinus: we could also use `python -c`, see https://github.com/dart-lang/dart-docker/issues/58#issuecomment-3223683013

src/backend/messages/urls.py
Nastaliss: Agreed

Nastaliss: Done in https://github.com/suitenumerique/messages/pull/324/commits/ca19dd9f99004b9977683cf430fe26b7e56aa3b9

src/backend/messages/wsgi.py
Nastaliss: Done in https://github.com/suitenumerique/messages/pull/324/commits/ca19dd9f99004b9977683cf430fe26b7e56aa3b9

src/backend/Dockerfile
Nastaliss: Didn't think of that, nice trick

Nastaliss: Done in https://github.com/suitenumerique/messages/pull/324/commits/a7c1d7d1ffc86e05168addea6e9eb4288e4dc4dd

PR #337, opened 2025-09-19
Blob size compressed

sdemagny:
## Summary by CodeRabbit

* **New Features**
  * Full message template/signature management (CRUD + render) with HTML/text previews; signatures can be selected in composer, applied to drafts, and included in sent messages.
  * New message composer and admin signature composer UI, inline signature block, variable insertion, and admin signatures page with data grid and compose modal.

* **Improvements**
  * Admin listings show additional blob info (including compressed size); translations updated; Toaster placement adjusted.

* **Bug Fixes**
  * Message body sanitization now updates when content changes.


src/backend/core/migrations/0008_blob_size_compressed.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 5.1.12 on 2025-09-19 11:54
+
+from django.db import migrations, models
+
+
+def set_size_compressed(apps, schema_editor):
+    Blob = apps.get_model('core', 'Blob')
+    queryset = Blob.objects.only("id", "raw_content", "size_compressed")
+    for blob in queryset.iterator(chunk_size=1000):
```
sylvinus: will this load 1000 blob contents in memory at once? we're ok with some slowness here it will run only once

sdemagny: Yes ! It was an optimization suggestion from CodeRabbit. But it's probably overkill at this point.

sylvinus:
Merged as https://github.com/suitenumerique/messages/commit/3dc8588b2947662d758f93ddc512f327a80f4d70

PR #378, opened 2025-10-14
✨(images) implement cid: image support

sylvinus:
Parses the email and replaces cid: images with blob urls

## Summary by CodeRabbit

- New Features
  - Inline images in messages now render when referenced by Content-ID (CID), using the message’s attachments.
  - Message view passes attachments to the message body to enable embedded image rendering.
  - API now exposes a CID field on attachments to support inline images.
  - Content Security Policy updated to allow images from the API origin for safe inline image loading.

- Refactor
  - Centralized logic for determining the API origin, simplifying URL construction across the app.

### DIFF
--- a/src/frontend/src/features/layouts/components/thread-view/components/thread-message/message-body.tsx
+++ b/src/frontend/src/features/layouts/components/thread-view/components/thread-message/message-body.tsx
@@ -86,7 +86,7 @@ const MessageBody = ({ rawHtmlBody, rawTextBody, attachments = [] }: MessageBody
             FORBID_TAGS: ['script', 'object', 'iframe', 'embed', 'audio', 'video'],
             ADD_ATTR: ['target', 'rel'],
         });
-    }, [rawHtmlBody, rawTextBody]);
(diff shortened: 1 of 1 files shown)


src/frontend/src/features/layouts/components/thread-view/components/thread-message/index.tsx
```diff
@@ -241,6 +241,7 @@ export const ThreadMessage = forwardRef<HTMLElement, ThreadMessageProps>(
                 <MessageBody
                     rawTextBody={textBody}
                     rawHtmlBody={htmlBody}
+                    attachments={message.attachments}
                 />
                 <footer className="thread-message__footer">
```
jbpenrath, line 246: In the AttachmentList, I'll exclude attachments with cid.

sylvinus: @jbpenrath this should be done in the frontend and also in the backend so that we don't set has_attachments

jbpenrath: Indeed !

PR #383, opened 2025-10-16
✨(threadevent) add ThreadEvents and Webhooks

### LINKED ISSUE #188: Rich threads
New objects should be displayed in the thread, as an event log:
* private messages
* log of important actions
* AI summaries?

sylvinus:
Webhook is a more generic version of the widget endpoint
ThreadEvents are blocks that can be inserted into Threads (fixes #188).

Frontend is still prototype, we will iterate on a few event types before merging

Capture d’écran 2025-11-03 à 10 27 46

## Summary by CodeRabbit

* **New Features**
  * Webhook integration for inbound message delivery and thread-event creation
  * Thread Events system with REST endpoints and admin listing
  * Timeline UI: thread events rendered alongside messages, including iframe-rich content

* **Tests**
  * Extensive tests covering webhook auth, message ingestion, and thread-event CRUD/validation

* **Documentation**
  * API spec updated to include thread-event schemas and endpoints

* **Style**
  * New styles for thread-event display in the thread view

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -1138,6 +1138,14 @@ class ThreadEvent(BaseModel):
         related_name="thread_events",
         help_text=_("Channel that created this event"),
     )
+    message = models.ForeignKey(
+        "Message",
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+        related_name="thread_events",
+        help_text=_("Message linked to this event"),
+    )
     data = models.JSONField(_("data"), default=dict, blank=True)
 
     class Meta:
--- a/src/frontend/src/features/providers/mailbox.tsx
+++ b/src/frontend/src/features/providers/mailbox.tsx
@@ -35,7 +35,7 @@ type MailboxContextType = {
     mailboxes: readonly Mailbox[] | null;
     threads: PaginatedThreadList | null;
     messages: PaginatedMessageList | null;
-    timelineItems: TimelineItem[] | null;
+    threadItems: TimelineItem[] | null;
     selectedMailbox: Mailbox | null;
     selectedThread: Thread | null;
     unselectThread: () => void;
@@ -61,7 +61,7 @@ const MailboxContext = createContext<MailboxContextType>({
     mailboxes: null,
     threads: null,
     messages: null,
-    timelineItems: null,
+    threadItems: null,
     selectedMailbox: null,
     selectedThread: null,
     loadNextThreads: async () => {},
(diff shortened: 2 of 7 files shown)


src/backend/core/models.py
```diff
@@ -1123,6 +1123,33 @@ def __str__(self):
...
+    class Meta:
+        db_table = "messages_threadevent"
+        verbose_name = _("thread event")
+        verbose_name_plural = _("thread events")
+        ordering = ["created_at"]
+
+    def __str__(self):
+        return f"{self.thread} - {self.type}"
+
+
```
jbpenrath, line 1160: In case I would like to create an event which relies on a specific message ? (e.g: Ask a review), the `message_id` will be stored into the `data` payload ?
I feel this model will have lot of entities so if we need to lookup on some fields those ones should be indexed and easy to access ?

src/frontend/src/features/providers/mailbox.tsx
```diff
@@ -180,6 +189,50 @@ export const MailboxProvider = ({ children }: PropsWithChildren) => {
...
+        }
+
+        // Sort by created_at
+        return items.sort((a, b) => {
+            const dateA = new Date(a.created_at).getTime();
+            const dateB = new Date(b.created_at).getTime();
+            return dateA - dateB;
+        });
+    }, [messagesQuery.data?.data?.results, threadEventsQuery.data?.data?.results, selectedThread]);
+
```
jbpenrath, line 235: Oh... by reading that I realize both message and thread event endpoints are paginated but we don't use an InfiniteQuery to fetch those resources so currently we always fetch only the first page...
Can we imagine to remove pagination from those resources (Is it acceptable from a performance point of view ?) ? It will ease the logic to merge data and ease its display.

src/frontend/src/features/providers/mailbox.tsx
```diff
@@ -311,6 +368,8 @@ export const MailboxProvider = ({ children }: PropsWithChildren) => {
         mailboxQuery,
         threadsQuery,
         messagesQuery,
+        threadEventsQuery,
+        timelineItems,
```
jbpenrath: `threadItems` ? Instead of `timelineItems` ?

src/frontend/src/features/providers/mailbox.tsx
sylvinus: up to you. It would make sense once it's more mature to merge all that in a single api call for a thread. I guess there will be huge threads so paginating would be a plus even if we put the default at ~100 messages.

PR #389, opened 2025-10-20
✨(template) insert template into message

sdemagny:
Capture d’écran 2025-10-21 à 01 15 36

Capture d’écran 2025-10-21 à 01 17 48

- [x] check placeholder name
- [ ] add feature flag FEATURE_MESSAGE_TEMPLATES

## Summary by CodeRabbit

* **New Features**
  * Insert pre-defined message templates from a new toolbar selector in the composer; modal supports loading/empty states and inserts at cursor or replaces an empty block, optionally adding a signature.
* **Localization**
  * Added template UI translations for English and French; other locale files updated with refreshed metadata (no translation text changes).
* **Style**
  * New styling for the template list and integrated styles into the editor.
* **Chores**
  * Updated translation file metadata and source-location references across locales.


src/frontend/src/features/blocknote/message-template-block/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+                    ...templateSignature,
+                    props: {
+                        ...templateSignature.props,
+                        mailboxId
+                    }
+                } as PartialMessageComposerBlockSchema);
+            }
+
+            // Insert blocks at cursor position
+            const currentBlock = editor.getTextCursorPosition().block;
```
jbpenrath, line 67: Select a template should replace the whole existing content nope?

src/frontend/src/features/blocknote/message-template-block/_index.scss
```diff
@@ -0,0 +1,46 @@
+.template-list {
+    display: flex;
+    flex-direction: column;
+    gap: 0.5rem;
+    padding: 1rem;
+    max-height: 400px;
+    overflow-y: auto;
```
jbpenrath: I don't think you need to set `overflow-y`. Furthermore, it seems to that, on Windows OS, setting `overflow-y` will always display the scroll bar.

src/frontend/src/features/blocknote/message-template-block/_index.scss
jbpenrath, line 1: Try to use as most as possible spacings token from cunningham to set `gap`, `padding` and `margin`
https://github.com/suitenumerique/messages/blob/c3ce9764d1362d37d52768ad7d63febc89aef1c4/src/frontend/src/styles/cunningham-tokens.css#L132-L156

jbpenrath, line 1: Where does those css properties (`--border-default-grey`, `--background-default-grey`, ...) come from ? You should use cunningham tokens
https://github.com/suitenumerique/st-messages/blob/c3ce9764d1362d37d52768ad7d63febc89aef1c4/src/frontend/src/styles/cunningham-tokens.css

src/frontend/src/features/blocknote/message-template-block/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+    if (templates.length === 0) {
+        return (
+            <Components.FormattingToolbar.Button
+                icon={<Icon name="description" size={IconSize.SMALL} />}
+                isDisabled={true}
+                label={t("No templates available")}
+                mainTooltip={t("No templates available")}
+            />
+        );
+    }
```
jbpenrath, line 103: What do you think to simply hide the button if there is no template ?

sdemagny: I figured that keeping the icon visible could encourage users to engage with the feature and help them better understand its purpose

src/frontend/src/features/blocknote/message-template-block/index.tsx
sdemagny: In my view, selecting a template should not overwrite existing content but instead insert content at the chosen location. Templates should be additive.

PR #392, opened 2025-10-26
Placeholder recipient name 

sdemagny:
Capture d’écran 2025-10-26 à 16 24 33

Capture d’écran 2025-10-26 à 16 15 34
Capture d’écran 2025-10-26 à 16 16 03

## Summary by CodeRabbit

* **New Features**
  * Templates can accept arbitrary query parameters (wildcard) as rendering context for more dynamic content.
  * Message composer auto-passes recipient info to templates for personalization.
  * Message template selector accepts an optional context prop for rendering; existing behavior remains compatible.

* **Localization**
  * Added "Recipient name" placeholder translations across supported locales.

* **Tests**
  * Added tests verifying template rendering with extended context (including custom attributes).


src/backend/core/models.py
```diff
@@ -1822,7 +1822,7 @@ def raw_body(self):
             return None
 
     def render_template(
-        self, mailbox: Mailbox = None, user: User = None
+        self, mailbox: Mailbox = None, user: User = None, context: Dict[str, Any] = None
```
jbpenrath:
```suggestion
        self, mailbox: Mailbox = None, user: User = None, context: Dict[str, Any] = {}
```

src/backend/core/models.py
```diff
@@ -1834,12 +1834,12 @@ def render_template(
         Returns:
             Dictionary with 'html_body' and 'text_body' keys containing rendered content
         """
-        name = (
+        context = context or {}
```
jbpenrath:
```suggestion
        context = context.copy()
```

PR #394, opened 2025-10-27
✨(attachments) configurable size limits with locale-aware formatting

mosa-riel:
Add DATA_UPLOAD_MAX_MEMORY_SIZE to settings.py to allow configuration via environment variable. Django's default is 2.5MB which is too small for emails with larger attachments.

Without this setting in settings.py, Django ignores the environment variable and uses its hardcoded default, causing RequestDataTooBig errors when the MTA tries to deliver emails exceeding 2.5MB.

Can be overridden via DATA_UPLOAD_MAX_MEMORY_SIZE environment variable.

## Summary by CodeRabbit

* **New Features**
  * Public config now exposes outgoing body and attachment size limits; in-memory upload limit auto-adjusts.

* **Bug Fixes / Validation**
  * Inbound emails over limit return 413 with clear error.
  * Outgoing/draft flows enforce body and cumulative attachment limits with localized validation errors.

* **UI**
  * Attachment uploader enforces dynamic cumulative limits, shows informational modals for rejections, and uses binary size formatting.
  * Removed text-loader skeleton visuals.

* **Localization**
  * Dutch locale removed; multiple translation updates and key changes across locales.

* **Tests**
  * Added tests for inbound and draft/attachment size limits and size-formatting.


sylvinus:
Good catch and I can confirm we have the issue on our instance too with medium-size incoming attachments. However I think it would be cleaner to have a single setting for the max attachment size (or 2, for incoming + outgoing?) and set that limit in accordance?

mosa-riel:
Yes,

I suggest we are going to use MAX_INCOMING_EMAIL_SIZE and MAX_OUTGOING_EMAIL_SIZE because I suppose we want to use the email body limited as well (gmail, for example, can do 50Mb email, and max 25mb attachment size).

On incoming mail, it is quite easy to get the MAX_INCOMING_EMAIL_SIZE and return 413 when the message is too large.

For outgoing mail, it is a UX question: I think it is best when creating a draft. But it is hard to determine the message body size (because it is not yet a raw email and updates very often when typing or pasting images).  We can easy calculate the total attachment size when uploading, so when an attachment exceeds that total size, we can show a message. 


When sending the final email, we can check the total size INCLUDING the body, but it can happen your email body is too large then. Do we want to check this also when **drafting** your email (harder to code) or is a message 'your total message is too large' when you try to **send** it is good enough?

mosa-riel:
WIP ;)

Long story somewhat short:

**init**
- attachment limits set on the backend .env
- backend communicates these to frontend (/api/v1.0/config)

**frontend**
- when uploading attachment frontend calculates side to prevent uploading the file to server
- when sending draft updates, backend checks message body size (works, pasted 5+ mb of text) and if exceeds, sends error and message to reduce body size

**backend**
- server also checks message and attechment size on send (for future API purposes or someone circumvents frontend)
- size communication is in MB (bytes / 1024 / 1024) because the shown size did not always correspond with the size my filesystem showed. The Intl comment of Coderabbit I understand.

mosa-riel:
thanks! thats clear.

Op ma 17 nov 2025, 11:07 schreef Sabrina Demagny ***@***.***>:

src/frontend/src/features/utils/attachment-helper/index.ts
```diff
@@ -67,14 +67,29 @@ export class AttachmentHelper {
...
-            unit: "byte",
-            unitDisplay: "narrow",
-          });
+        // Use binary (1024) conversion for user-friendly display
+        // This matches file system conventions and gives clean numbers
+        const formatter = (value: number, unit: string) => {
+            return new Intl.NumberFormat(language, {
+                style: 'unit',
+                unit: unit,
+                maximumFractionDigits: 1
```
sdemagny, line 77: Nitpick: Could you keep unitDisplay: "narrow"?

src/backend/core/tests/api/test_attachments.py
```diff
@@ -336,3 +336,293 @@ def test_add_attachment_to_existing_draft_and_send(
         assert parts[4].get_payload(decode=True).decode() == blob.get_content().decode()
         assert parts[4].get_content_disposition() == "attachment"
         assert parts[4].get_filename() == "test_attachment.txt"
+
+    def test_draft_attachment_size_limit_exceeded(self, api_client, user_mailbox):
+        """Test that adding attachments exceeding the size limit raises ValidationError."""
+        from django.test import override_settings
```
sdemagny: Please move this import to the top and remove all the other imports

sdemagny (changes requested):
Some minor returns in addition to removing the modal and cleaning up the OpenAPI schema ;) 
Thank you very much for this PR. I’d love to merge it as soon as possible, if that works for you. How would you prefer to proceed? Would you like to make the adjustments yourself, or would you like me to take your code in a PR on our side and implement my feedback?

sdemagny:
@mosa-riel Thanks! Everything looks good :) Could you please rebase/resolve the conflicts and squash the commits?

src/backend/core/api/permissions.py
```diff
@@ -148,20 +148,31 @@ def has_object_permission(self, request, view, obj):
...
-                    enums.MailboxRoleChoices.EDITOR
-                    if view.action == "destroy"
-                    else enums.MailboxRoleChoices.SENDER
-                )
-                return thread.accesses.filter(
-                    role=enums.ThreadAccessRoleChoices.EDITOR,
-                    mailbox__accesses__user=user,
-                    mailbox__accesses__role__gte=minimal_user_role,
-                ).exists()
+                mailbox = thread.accesses.get(mailbox__accesses__user=user).mailbox
```
sylvinus, line 151: This looks like it reverts https://github.com/suitenumerique/messages/commit/f2260a177ac350c0c55022549503b9d3acf0cead ?

PR #398, opened 2025-10-29
✨(feature): Image proxy for CSP compliance and privacy protection

### LINKED ISSUE #187: Images in emails
Right now all images are blocked with the CSP.

We should support both external images (proxified, maybe approved by the user) and internal (cid:) images.

mosa-riel:
## Summary

This PR adds an image proxy feature that enables strict Content Security Policy (CSP) enforcement while maintaining email functionality, and provides privacy protection by proxying external images through the application server.

Closes #187

## Motivation

**Primary goal: CSP Compliance**

Modern web applications need strict Content Security Policy headers to prevent XSS attacks. However, emails often contain external images that violate a strict `img-src 'self'` CSP policy.

By proxying all external images through the application server, all image requests originate from the same domain, enabling:
```
Content-Security-Policy: img-src 'self'
```

**Secondary goal: Privacy Protection**

External images in emails are commonly used as tracking pixels. The image proxy prevents exposure of user IP addresses and reading behavior to third-party servers.

## Security

The image proxy implements defense-in-depth security measures to prevent Server-Side Request Forgery (SSRF) attacks:

**SSRF Protection:**
- URL validation before making any HTTP request
- Blocks all IP addresses (only domain names allowed)
- DNS resolution validation to prevent DNS rebinding attacks
- Blocks private IP ranges, loopback, link-local, multicast, and cloud metadata endpoints
- Returns informative SVG placeholder for blocked images

**Memory Protection:**
- Streams content in 8KB chunks to prevent memory exhaustion
- Enforces configurable size limits (default 5MB)

**Content Validation:**
- Validates content is actually an image using libmagic
- Verifies Content-Type headers
- Returns SVG placeholder for invalid content

## Implementation

**Backend:**
- New API endpoint: `GET /api/v1.0/mailboxes/{mailbox_id}/image-proxy/?url={encoded_url}`
- SSRF protection utility in `core/utils.py`
- HTML rewriting in `MessageSerializer` to proxy external images and handle CID attachments
- Configurable via `PROXY_EXTERNAL_IMAGES` and `PROXY_MAX_IMAGE_SIZE_MB` environment variables
- 30-day cache headers on blob downloads

**Frontend:**
- Lazy loading on all images for improved page load performance
- HTTP/2 enabled in nginx for parallel image requests

**Dependencies:**
- Added `beautifulsoup4` for HTML parsing
- Uses `python-magic` for MIME type validation

## URL rewriting

**External images:**
```html
<!-- Before (violates CSP) -->
<img src="https://example.com/image.jpg">

<!-- After (CSP compliant) -->
<img src="/api/v1.0/mailboxes/123/image-proxy/?url=..." loading="lazy">
```

**CID attachments:**
```html
<!-- Before -->
<img src="cid:image001@example.com">

<!-- After -->
<img src="/api/v1.0/blob/msg_456_0/download/" loading="lazy">
```

## Benefits

- **CSP Compliance**: Enables strict `img-src 'self'` security policy
- **Privacy Protection**: Prevents IP address leakage and tracking
- **SSRF Protection**: Comprehensive validation prevents internal network access
- **Performance**: Lazy loading + HTTP/2 + caching improve load times
- **Configurable**: Enable per deployment with adjustable size limits

## Configuration

```bash
# Enable image proxying (default: false)
PROXY_EXTERNAL_IMAGES=true

# Maximum image size in MB (default: 5)
PROXY_MAX_IMAGE_SIZE_MB=10
```

## Summary by CodeRabbit

* **New Features**
  * External images in messages can now be proxied through a secure endpoint with configurable size limits.
  * Images in message threads now load with lazy loading.

* **Improvements**
  * Blob downloads now cache for 30 days in the browser.

### DIFF
--- a/src/backend/core/api/viewsets/image_proxy.py
+++ b/src/backend/core/api/viewsets/image_proxy.py
@@ -1,9 +1,9 @@
 """API ViewSet for proxying external images."""
 
-import imghdr
 import logging
 from urllib.parse import unquote
 
+import magic
 import requests
 from django.conf import settings
 from django.http import HttpResponse
@@ -165,9 +165,9 @@ class ImageProxyViewSet(ViewSet):
             image_content = b"".join(chunks)
 
             # Validate that content is actually an image (defense in depth)
-            image_type = imghdr.what(None, h=image_content)
-            if not image_type:
-                logger.warning("Content from %s is not a valid image", url)
+            mime_type = magic.from_buffer(image_content, mime=True)
+            if not mime_type.startswith("image/"):
+                logger.warning("Content from %s is not a valid image: %s", url, mime_type)
                 # Return placeholder image for invalid content
                 svg_placeholder = """<svg xmlns="http://www.w3.org/2000/svg" width="400" height="100" viewBox="0 0 400 100">
   <rect width="100%" height="100%" fill="#f8f9fa"/>


mosa-riel:
@coderabbitai review

jbpenrath (commented):
Nice job ❤️

Can you add unit tests for backend utils and new endpoint ?

mosa-riel:
Good comments, thank you for the insights.  I created this so I have a working version for here, but I would be happy to integrate more of your comments.


_I think we need to add some level of user consent to load external images. Even though the user IP can't leak because we proxy the request, many external images contain individualized tracking links that can report that an email was read._

That is true. I don't think that is hard to implement. Let me update the PR later.

_I'm not sure we should parse the email content to rewrite the image srcs on the backend. We already have a parsing hook on the frontend (modified in this patch), it would be a good place to implement the rewrite instead? This could then be done dynamically based on user consent._

What parsing hook are  you referring too? I don't really have a preference here, I thought it was nice to not give any external URL to the frontend at all. The consent could just trigger a new fetch from the backend with a tag to add the images. 

_we could also parse attributes like image dimensions (to block trackers that are 1px for instance)._
This would also be nice to add same dimension placeholders to prevent mailformatting to be skewed ?

_isn't image/svg+xml a vector for injecting scripts or other nasty stuff? It would probably be better to have a stricter mime type whitelist_
Yes, I am not 100% aware of this but the reason it is  in a separate file is so it is somewhat expandable.

I will take these remarks and update the PR.

@jbpenrath any other remarks?

PR #402, opened 2025-10-31
(socks-proxy) add healthcheck to dockerfile

zeylos:
## Purpose

Description...

## Proposal

Description...

- [] item 1...
- [] item 2...

## Summary by CodeRabbit

* **Chores**
  * Added automated runtime health checks for the SOCKS proxy.
  * Hardened build and startup processes for more reliable container operation (noninteractive package installs and runtime tooling ensured).
  * Relaxed strict file-permission handling in favor of a safer startup umask to better support runtime updates and secret handling.

### DIFF
--- a/src/socks-proxy/Dockerfile
+++ b/src/socks-proxy/Dockerfile
@@ -6,7 +6,7 @@ ARG DANTE_SHA256=1973c7732f1f9f0a4c0ccf2c1ce462c7c25060b25643ea90f9b98f53a813fae
 
 RUN <<EOR
 apt-get update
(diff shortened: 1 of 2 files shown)


zeylos:
Netcat was generating way too much logs, and I didn't find any easy way to delete netcat logs and keep the actual traffic's logs. the current healthcheck is not actually that useful - it only checks if the process is not in Disk Sleep or Zombie state. This should do, we can work on it if we find any weird edge case later

sylvinus (approved):

PR #419, opened 2025-11-11
🩹(mime) fix inline HTML being treated as attachment

mosa-riel:
## Problem

HTML email bodies with `Content-Disposition: inline` were being incorrectly classified as downloadable attachments instead of being rendered as the message body.

## Cause

The MIME parser treated ANY `Content-Disposition` header (including `inline`) as an attachment indicator. This violated RFC 2183 which distinguishes between:
- `Content-Disposition: inline` - should be displayed automatically (body content)
- `Content-Disposition: attachment` - should be presented as a downloadable file

## Solution

Updated the classification logic in `parser.py:258-267`:
1. Always treat `Content-Disposition: attachment` as attachments
2. Treat `Content-Disposition: inline` with `text/plain` or `text/html` as message body
3. Keep existing behavior for parts with filenames but no disposition

## Testing

To reproduce the bug, send this email to your mailbox:

```
From: sender@example.com
To: your-mailbox@example.com
Subject: Test inline HTML
MIME-Version: 1.0
Content-Type: text/html; charset=UTF-8
Content-Disposition: inline

<html><body><p>This should be the message body, not an attachment.</p></body></html>
```

**Before:** Message appears with empty body and "unnamed" attachment (text/html)  
**After:** HTML is rendered as the message body

Fixes issue where auto-reply emails from Gmail were showing as attachments.

## Summary by CodeRabbit

* **Refactor**
  * Improved email parsing to more accurately classify attachments, inline parts, and message bodies.
  * Attachments consistently include a filename (or "unnamed"), binary content and a content hash for reliable integrity checks.
  * Text/plain and text/html bodies are preserved and decoded into the appropriate body fields.
  * Content-type handling and logging simplified for clearer, more consistent message handling.

### DIFF
--- a/src/backend/core/mda/rfc5322/parser.py
+++ b/src/backend/core/mda/rfc5322/parser.py
@@ -212,10 +212,8 @@ def parse_message_content(message) -> Dict[str, Any]:
 
         # 2. If not found via Flanker property, try parsing raw headers
         if not filename:
-            # headers_dict is already defined above
-
             # 2a. Try Content-Disposition header parsing (regex method)
-            # disposition_header is already defined above
+            disposition_header = headers_dict.get("Content-Disposition", "")
             if disposition_header and "filename=" in str(disposition_header):
                 match_disp = re.search(
                     r'filename\*?=(?:(["\'])(.*?)\1|([^;]+))',


mosa-riel:
I am not really convinced.

Why can't we use Flanker's build-in methods?

E.g. I see issues when a calendar invite is attached, like this mail:

```
From: sender@example.com
To: recipient@example.com
Subject: Meeting Invitation
Date: Mon, 11 Nov 2025 12:00:00 +0000
Message-ID: <test123@example.com>
MIME-Version: 1.0
Content-Type: multipart/alternative;
	boundary="----=_Part_123_456.789"

------=_Part_123_456.789
Content-Type: text/plain; charset="utf-8"

This is a meeting invitation.

------=_Part_123_456.789
Content-Type: text/calendar; charset="utf-8"; method=REQUEST

BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//Example Corp//Calendar//EN
METHOD:REQUEST
BEGIN:VEVENT
UID:meeting-123@example.com
DTSTART:20251115T140000Z
DTEND:20251115T150000Z
SUMMARY:Team Meeting
DESCRIPTION:Discuss project updates
ORGANIZER:mailto:sender@example.com
ATTENDEE:mailto:recipient@example.com
END:VEVENT
END:VCALENDAR

------=_Part_123_456.789--
```

 The text/calendar part has no Content-Disposition header and no filename parameter. According to the standards and how major email clients handle it, this should be displayed inline (not as a downloadable attachment).

The current parser is falling through to the "treat other string content as attachment" case, which is incorrect.


This would fix it?
  
```
  if part.is_attachment():
      # This part has Content-Disposition: attachment
      # Treat as downloadable attachment
  elif part.is_body():
      # This part has no filename AND is text/* or message/*
      # Treat as message body (text/plain, text/html, text/calendar, etc.)
  elif part.is_inline():
      # This part has Content-Disposition: inline
      # Could be inline images (with CID), or inline calendar invites
      # Still store as attachment but with disposition="inline"
  else:
      # Fallback for other parts
```

mosa-riel:
The flankers methods make more sense and they work well in our environment.

sylvinus:
@mosa-riel we'd like to merge this next, could you send a clean PR? I think there are again unrelated commits here

PR #440, opened 2025-12-05
Replace hardcoded Fichiers strings with dynamic Drive app name variable

NielsCodes:
## Purpose

The Drive file picker currently has the "Fichiers" name for the Drive app hardcoded in the translation strings.
Fichiers already uses the name "Drive" in other contexts and other implementations of the suite might use a different name altogether.

Screenshot 2025-12-05 at 12 09 26

## Proposal

- Replace the hardcoded "Fichiers" in the translation files with a variable
- Allow setting of `DRIVE_APP_NAME` environment variable to override the name (defaults to Drive)
- Add the Drive app name to the config endpoint so it can be passed from the backend to the frontend

Screenshot 2025-12-05 at 12 09 54

**Note:** This still assumes the same name will be used across all languages.

## Summary by CodeRabbit

* **New Features**
  * Drive application name is configurable and used dynamically in tooltips, labels, status messages and previews across locales (defaults to "Drive").

* **API**
  * Public configuration now includes the Drive app name so clients receive and display the configured name.

* **Behavior**
  * Drive preview and feature availability now respect feature flag and DRIVE configuration (shown only when properly configured).

* **Docs**
  * New environment variable documented to set the Drive app name.

* **Tests**
  * Config tests updated to include the new Drive app name field.


src/backend/messages/settings.py
```diff
@@ -781,6 +781,9 @@ class Base(Configuration):
         "sdk_url": "/sdk",
         "api_url": "/api/v1.0",
     }
+    DRIVE_APP_NAME = values.Value(
+        default="Drive", environ_name="DRIVE_APP_NAME", environ_prefix=None
+    )
```
jbpenrath: What do you think to put the `app_name` within the `DRIVE_CONFIG` dict ?
```suggestion
DRIVE_CONFIG = {
"base_url": values.Value(
default=None, environ_name="DRIVE_BASE_URL", environ_prefix=None
),
"file_url": "/explorer/items/files",
"sdk_url": "/sdk",
"api_url": "/api/v1.0",
"app_name": values.Value(
        default="Drive", environ_name="DRIVE_APP_NAME", environ_prefix=None
    )
}
```

src/backend/core/api/viewsets/config.py
```diff
@@ -137,6 +141,7 @@ def get(self, request):
                         "sdk_url": f"{base_url}{settings.DRIVE_CONFIG.get('sdk_url')}",
                         "api_url": f"{base_url}{settings.DRIVE_CONFIG.get('api_url')}",
                         "file_url": f"{base_url}{settings.DRIVE_CONFIG.get('file_url')}",
+                        "app_name": settings.DRIVE_APP_NAME,
```
jbpenrath: According to my comment into `settings.py`


```suggestion
                        "app_name": settings.DRIVE_CONFIG.get('app_name'),
```

PR #441, opened 2025-12-05
Add dutch patterns for attachment detection

NielsCodes:
## Purpose

The patterns that trigger an email attachment warning are currently only defined in English and French. 

## Proposal

Add detection patterns for Dutch. 🇳🇱🧀

## Summary by CodeRabbit

* **Chores**
  * Added Dutch localization for attachment-detection phrases, covering keywords, abbreviations, and varied expressions to improve Dutch-language support.
  * Standardized French apostrophe patterns in attachment-detection phrases to improve consistency and reliability of French-language processing.

### DIFF
--- a/src/frontend/src/features/i18n/attachments-detection-map.json
+++ b/src/frontend/src/features/i18n/attachments-detection-map.json
@@ -98,9 +98,7 @@
       "bijgevoegd document",
       "ingesloten",
       "zie bijlage",
-      "in de bijlage",
-      "bijgevoegd is",
-      "bijgevoegd zijn"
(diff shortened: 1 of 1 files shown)


MathisRouget:
> @MathisRouget I noticed that the detection only uses an `includes()` check, so I guess the dynamic patterns would never be triggered (unless someone literally puts the entire pattern in their email lol).
> 
> Is my assumption correct, or did I miss some part of the pattern match?

Good catch, when I added this feature, we did not use dynamic patterns and indeed with them, the `includes` method will never return `True`.

PR #454, opened 2025-12-11
👔(backend) add flag `is_trashed` to thread resource

jbpenrath:
## Purpose

Add a flag `is_trashed` to Thread model in order to monitor when a thread is fully trashed. It helps to exclude those kind of threads in some views with inactive message like archive or spam for example.

**Then by default we exclude fully trashed and spam threads from results until user is looking for trashed messages.**

## Summary by CodeRabbit

* **New Features**
  * Threads now expose a read-only "trashed" status when all messages in a thread are trashed
  * New query parameter to filter threads by trashed status

* **Changed**
  * Thread listings and search exclude trashed and spam threads by default (overridable)
  * Public API/schema updated to include trashed status and clarify related filter descriptions

* **Tests**
  * Test fixtures and expectations updated to reflect exclusion of trashed/spam threads


sylvinus (approved):
LGTM

PR #432, opened 2025-11-21
🩹(mime) parsing fixes

mosa-riel:
## Summary

  Fix MIME parsing issues where inline HTML and calendar invites were incorrectly treated as downloadable attachments, and bodytext as unnamed attachments.

  Cherry-picked from upstream PR #419 (suitenumerique/messages).

  ## Changes

  - Fix inline HTML being treated as attachment
  - Refactor parser to use Flanker's built-in `is_attachment()`, `is_body()`, and `is_inline()` methods
  - Fix text/calendar parts with `Content-Disposition: inline` being misclassified
  
  ## Commits

  - 🩹(mime) fix inline HTML being treated as attachment
  - ♻️(mime) refactor parser to use Flanker's built-in methods
  - fix: Add missing disposition_header variable definition

## Summary by CodeRabbit

* **Refactor**
  * Improved email parsing to use library-driven part classification, streamlining identification of attachments, inline content, and message bodies.
  * Consolidated and simplified handling so text/HTML bodies and attachments are processed more reliably.

* **Bug Fixes**
  * Better filename inference for attachments (adds typical extensions when missing) so attachments and inline images get sensible names.
  * More robust fallback and error handling for unclassified parts, reducing dropped or misclassified content.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #463, opened 2025-12-26
🐛(rfc5322) improve email parsing & display multiparts in frontend

sylvinus:
This builds upon #432 and adds :
 - tests
 - attachment filename sanitization
 - better multipart parsing logic, adapted from [section 4.1 of the JMAP spec](https://jmap.io/spec-mail.html#properties-of-the-email-object). This should fix some display issues with inlined images from Apple Mail for instance.
 - rendering of multipart messages in frontend

## Summary by CodeRabbit

* **Bug Fixes & Improvements**
  * Improved email parsing: better multipart handling, clearer inline vs. attachment classification, safer filename inference/sanitization, attachment hashing, and simplified parse error messaging.

* **Frontend**
  * Message rendering now consumes per-part bodyParts, with per-part renderers and CID→blob URL support for embedded images; unified HTML/text rendering.

* **API**
  * New MessageBodyItem schema for richer body-part metadata surfaced in the API.

* **Tests**
  * Greatly expanded parsing tests covering encodings, dates, nested multiparts, filenames, labels, and edge cases.

* **Chores**
  * Updated IMAP domain regex patterns.


src/backend/core/mda/rfc5322/parser.py
```diff
@@ -139,230 +141,394 @@ def parse_date(date_str: str) -> Optional[datetime]:
...
+
+    filename = filename.strip('"/.\\')
+
+    # Remove null bytes and control characters
+    filename = re.sub(r"[\x00-\x1f\x7f]", "", filename)
+
+    # Remove dangerous characters
+    filename = re.sub(r'[<>:"|?*\\/]', "_", filename)
+
+    return filename[:max_length]
```
jbpenrath: It could truncate file extension.

src/frontend/src/features/layouts/components/thread-view/components/thread-message/index.tsx
```diff
@@ -62,8 +61,40 @@ export const ThreadMessage = forwardRef<HTMLElement, ThreadMessageProps>(
...
+        const [processedHtmlBody, driveAttachments] = useMemo(() => {
+            if (message.htmlBody.length === 0) {
+                return [[], []] as const;
+            }
+            // Process each HTML body part for drive attachments
+            const allDriveAttachments: ReturnType<typeof MailHelper.extractDriveAttachmentsFromHtmlBody>[1] = [];
+            const processedParts = message.htmlBody.map(part => {
+                const partContent = (part as { content?: string }).content ?? "";
+                const partType = (part as { type?: string }).type ?? "text/html";
+                const partId = (part as { partId?: string }).partId ?? "";
```
jbpenrath: If you need to do that seems there is an issue about the openapi schema. (Take a look to my fixup commit)

src/frontend/src/features/layouts/components/thread-view/components/thread-message/renderers/index.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Render all body parts to a single HTML string, with separators between parts.
+ */
+export function renderBodyParts(
+    parts: readonly BodyPart[],
+    cidToBlobUrlMap: Map<string, string>
+): string {
+    return parts
+        .map((part) => renderBodyPart(part, cidToBlobUrlMap))
+        .filter(Boolean)
+        .join("<br/><br/>");
```
jbpenrath: IMO this is not a good idea to modify html structure by adding blank lines.
If we have some issues to display content types we should use css

src/frontend/src/features/layouts/components/thread-view/components/thread-message/renderers/text_plain.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Render a text/plain body part to HTML.
+ * Escapes the content using browser-native textContent for safe XSS prevention.
+ */
+export function renderTextPlain(content: string): string {
+    if (!content) {
+        return "";
+    }
+    const tempDiv = document.createElement("div");
+    tempDiv.textContent = content;
+    return `<div class="text-plain-content">${tempDiv.innerHTML}</div>`;
```
jbpenrath: What do you think to render a paragraph as this is plain text content.


```suggestion
    const paragraph = document.createElement("p");
    paragraph.textContent = content;
    paragraph.classList.add('text-plain-content')
    return paragraph.outerHTML();
```

src/frontend/src/features/layouts/components/thread-view/components/thread-message/index.tsx
```diff
@@ -62,8 +62,41 @@ export const ThreadMessage = forwardRef<HTMLElement, ThreadMessageProps>(
...
+        // Extract drive attachments from HTML body parts
+        const [processedHtmlBody, driveAttachments] = useMemo(() => {
+            if (message.htmlBody.length === 0) {
+                return [[], []] as const;
+            }
+            // Process each HTML body part for drive attachments
+            const allDriveAttachments: ReturnType<typeof MailHelper.extractDriveAttachmentsFromHtmlBody>[1] = [];
+            const processedParts = message.htmlBody.map(part => {
+                const partContent = part?.content ?? "";
+                const partType = part?.type ?? "text/html";
```
jbpenrath: According to types `part.type` is always a string. Is that true ? If so, the nullish operator `??` will never work you should use `||` instead to get a fallback if the string is empty.

```js
// part.type = ""
const partType = part?.type ?? "text/html"; // == ""
const partType = part?.type || "text/html"; // == "text/html"
```

jbpenrath (approved):
LGTM 🧩

PR #469, opened 2026-01-07
✨(all) allow to display external images through proxy

jbpenrath:
## Purpose

Allow users to display external images from their email through a secure proxy endpoint to ensure security and privacy and respect iframe csp policy.

https://github.com/user-attachments/assets/ba47c8ae-2a10-4cf1-9051-828333afc58b

_Note: In the demo video, the three images with an exclamation mark are svg images that have been considered as suspicious content by the proxy._

Based on #398 

## Summary by CodeRabbit

* **New Features**
  * Opt-in mailbox image proxy with a public GET endpoint; IMAGE_PROXY_ENABLED exposed in config.

* **UI / Localization**
  * Per-message consent banner/action to display external images; added UI text and translations.

* **Behavior**
  * Proxied images served with SSRF protections, MIME and size checks, safe fallbacks, and long browser caching (~30 days).

* **Documentation**
  * Added environment docs for image proxy defaults, max size, and cache TTL.

* **Tests**
  * Comprehensive test coverage for proxy behavior, protections, validations, redirects, and error cases.


src/backend/core/api/viewsets/image_proxy.py
```diff
@@ -0,0 +1,466 @@
...
+                logger.warning(
+                    "Content from %s is not a valid image: %s", url, mime_type
+                )
+                # Return placeholder image for invalid content
+                return ImageProxySuspiciousResponse(
+                    status=http_status.HTTP_400_BAD_REQUEST
+                )
+
+            # Check that mime type is not a blacklisted image type
+            if mime_type in enums.BLACKLISTED_PROXY_IMAGE_MIME_TYPES:
```
sylvinus, line 411: I feel a whitelist would be more secure?

src/backend/core/api/viewsets/image_proxy.py
```diff
@@ -0,0 +1,466 @@
...
+                    {"error": "Image too large"},
+                    status=http_status.HTTP_413_REQUEST_ENTITY_TOO_LARGE,
+                )
+
+            return HttpResponse(
+                image_content,
+                content_type=mime_type,
+                headers={
+                    "Cache-Control": f"public, max-age={settings.IMAGE_PROXY_CACHE_TTL}",
+                    "X-Proxied-From": url,
```
sylvinus: What is the purpose of X-Proxied-From ?

src/backend/core/api/viewsets/image_proxy.py
```diff
@@ -0,0 +1,466 @@
...
+                )
+                return Response(
+                    {"error": "Image too large"},
+                    status=http_status.HTTP_413_REQUEST_ENTITY_TOO_LARGE,
+                )
+
+            return HttpResponse(
+                image_content,
+                content_type=mime_type,
+                headers={
```
sylvinus, line 451: We should add X-Content-Type-Options: nosniff https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Content-Type-Options

sylvinus: What about also content-disposition: attachment ?

sylvinus: Django seems to add nosniff with the securitymiddleware but it could be more explicit to also add it here manually. https://docs.djangoproject.com/en/6.0/ref/middleware/#x-content-type-options-nosniff

sylvinus: Also add a CSP ?

src/backend/core/api/viewsets/image_proxy.py
jbpenrath: I was just afraid by the maintenance of such a list but and the user support generated by a strict list.

WDYT to make this list overridable through env var ?
IMO a blacklist with a strict CSP is already secure and reduce the risk of maintenance ?

src/backend/core/api/viewsets/image_proxy.py
jbpenrath: Debug. I removed it to prevent some exploit by accessing original url.

src/backend/core/api/viewsets/image_proxy.py
jbpenrath: I add a strict CSP and also [Permission Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Permissions_Policy)
Instead to declare explicitly headers already added by Django, I prefer to write a test which checks that all security headers exists in the response. In this way we will be aware to identify regression later.

sylvinus (approved):
LGTM, nice work

PR #486, opened 2026-01-16
✨(storage) implement tiered storage

### LINKED ISSUE #185: Scalable email storage

sylvinus:
This allows to use S3-compatible object storage to offload blobs, making Postgres much lighter. We design for storing ~1B emails on a single instance.

Fixes #185.

## Release Notes

* **New Features**
  * Blob offloading to S3-compatible object storage with automatic hourly scheduling based on age and size thresholds
  * Optional AES-256-GCM encryption for blob data at rest with configurable key rotation
  * Automated garbage collection system for unreferenced blobs
  * Enhanced admin interface displaying blob storage location and encryption key assignment

* **Documentation**
  * New tiered storage architecture guide and deployment configuration instructions


src/backend/core/models.py
```diff
@@ -2114,13 +2124,21 @@ def create_blob(
                 {"compression": f"Unsupported compression type: {compression}"}
             )
 
+        # Encrypt content if encryption keys are configured
+        # pylint: disable-next=import-outside-toplevel
+        from core.services.tiered_storage import TieredStorageService
```
jbpenrath: I feel we can make this import global without cyclic import issue.

src/backend/core/models.py
```diff
@@ -2089,6 +2084,16 @@ def create_blob(
...
+        # Resolve algorithm + level from the configured spec ("zstd:3" etc.).
+        try:
+            default_algo, zstd_level = parse_compression_spec(
+                settings.MESSAGES_BLOB_COMPRESS
+            )
+        except ValueError as exc:
+            raise ValidationError({"compression": str(exc)}) from exc
+        if compression is None:
+            compression = default_algo
+
```
jbpenrath, line 2127: I did not see when `compression` arg is used in the codebase. I feel this strange to be able to pass compression algo from args but silently retrieve zstd compression level from settings. IMO if this args is not used, let's remove this logic.

jbpenrath (commented):
Just nitpicks

docs/tiered-storage.md
```diff
@@ -0,0 +1,142 @@
+# Tiered blob storage
+
+Blobs (raw RFC822 email bodies and attachments) live in PostgreSQL by
+default. Once a blob is older than `TIERED_STORAGE_OFFLOAD_AFTER_DAYS`,
+a periodic celery task moves its bytes to S3 and clears the PG row's
+`raw_content`. Reads transparently fetch from whichever location the
+row points at — application code only ever calls `blob.get_content()`.
+
```
chibenwa, line 8: Have you concidered moving message body onto the object store while keeping headers around for longer?

This is generally performance critical (no object store hit and no heavy mail parsing) on email list load, while actually it is OK to store headers around in DB for much longer (months?)

docs/tiered-storage.md
```diff
@@ -0,0 +1,142 @@
...
+## Architecture
+
+- **Storage path**: `blobs/{key_id}/{sha[:3]}/{sha}`. The leading
+  `key_id` segment lets blobs encrypted with different keys coexist
+  (essential for crash-safe online key rotation). The 3-char sha
+  prefix shards each key into 4096 sub-prefixes for S3 request-rate
+  balance.
+- **Deduplication**: blobs sharing the same SHA-256 share the same S3
+  object. The DB is the source of truth; the existence check on S3 is
+  a defensive guard against external deletions.
```
chibenwa: Nice but how is this GCed if data is deleted ?

docs/tiered-storage.md
sylvinus: Hi! We extract enough metadata from the headers to the DB (and to OpenSearch) to be able to have all email lists load without hitting the mime data, so I think it's already the case even if we don't store all headers?

docs/tiered-storage.md
sylvinus: We removed the existence check on S3 - external deletions on S3 are not a risk in the scope

src/backend/core/services/tiered_storage.py
```diff
@@ -0,0 +1,542 @@
...
+
+        def _run(s=sha256_bytes, k=key_id):
+            try:
+                self.delete_if_orphaned(s, k)
+            except Exception:  # pylint: disable=broad-except
+                logger.exception(
+                    "Post-commit S3 cleanup failed for sha=%s key_id=%d; "
+                    "verify_blobs --mode=storage-to-db will list strays",
+                    s.hex()[:8],
+                    k,
```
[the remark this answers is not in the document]
sylvinus: key_id is an integer, not sensitive.

docs/tiered-storage.md
chibenwa: Great, that will do the job actually I presume.

With all these data, what is the actual footprint of a typical message once tiered in the DB ?

sylvinus:
Merged as https://github.com/suitenumerique/messages/commit/f36057079810086d637e7eece26b5b1dac09927d

PR #498, opened 2026-01-21
🔧(backend) update Django admin URL handling

jbpenrath:
## Purpose

Modified the Django admin URL configuration to ensure it ends with a trailing slash. Updated related references in the Nginx configuration and environment variables for consistency.

✅ Tested on staging instance

## Summary by CodeRabbit

* **Documentation**
  * Added DJANGO_ADMIN_URL environment variable docs with default value and usage guidance (must not end with a trailing slash).

* **Changes**
  * Enforced no-trailing-slash requirement for the admin URL configuration.
  * Adjusted admin routing to remain compatible with the revised configuration format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Changed
+
+- Now `DJANGO_ADMIN_URL` must not end with `/`.
+
 ## [0.1.0] - 2026-01-20
 
 ### Added
--- a/docs/env.md
+++ b/docs/env.md
@@ -34,6 +34,7 @@ The application uses a new environment file structure with `.defaults` and `.loc
 | `DJANGO_SETTINGS_MODULE` | `messages.settings` | Django settings module | Required |
 | `DJANGO_SUPERUSER_PASSWORD` | `admin` | Default superuser password for development | Dev |
 | `DJANGO_DATA_DIR` | `/data` | Base directory for data storage | Optional |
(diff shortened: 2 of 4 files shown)


src/frontend/nginx/nginx.conf.template
```diff
@@ -19,11 +19,7 @@ server {
     }
 
     # Django admin
-    location /${DJANGO_ADMIN_URL} {
-        return 301 https://$host/${DJANGO_ADMIN_URL}/;
-    }
```
zeylos, line 24: ok for the scalingo part but we need to keep this part to redirect /admin to /admin/ ; given the fact that we come from the frontend and that the backend is not aware of the "real port", /admin redirects to `host.domain:3000/admin/` without this config.

If you really need the ending / in the var for something we may be able to do some tricks here to get the admin url without the last char or something, but if you don't we should let the thing like it was

jbpenrath: Yup if the var does not end with `/`, it mess up the backend routing logic. We turn into an infinite redirect loop when the user is not authentified and django redirects the user to the admin login page.
As the purpose of this var is to hide a little bit the admin access, I assumed to remove this redirect. But your right let's keep that and tweak it to work with the new var format.

**EDIT: So finally, [as it seems not as simple to fix the nginx redirect rule](https://github.com/suitenumerique/messages/pull/498#discussion_r2713940085), I revamp my PR and update the backend logic to work with env var without ending with a backslash**

zeylos (approved):

zeylos:
test preprod OK

PR #506, opened 2026-01-25
✨(throttle) add outbound message recipients throttling

sylvinus:
This is inspired by #466 but focuses on outbound backend-only changes.

## Release Notes

* **New Features**
  * Added outbound throttling for external recipients with configurable per-mailbox and per-domain rate limits
  * Expanded image proxy configuration with explicit settings for enabling, sizing, and caching
  * Added throttle status visibility in the admin interface

* **Documentation**
  * Updated environment variables documentation with new throttling and image proxy configuration options

* **Tests**
  * Added comprehensive test coverage for throttling functionality

### DIFF
--- a/src/backend/core/services/throttle.py
+++ b/src/backend/core/services/throttle.py
@@ -7,7 +7,7 @@ cache counters with fixed time windows.
 import logging
 import math
 from datetime import timedelta
-from typing import Any, Optional
+from typing import Any
 
 from django.conf import settings
 from django.core.cache import cache
--- a/src/backend/core/admin.py
+++ b/src/backend/core/admin.py
@@ -1,4 +1,5 @@
 """Admin classes and registrations for core app."""
+# pylint: disable=too-many-lines
 
 import logging
 


src/backend/core/services/throttle.py
```diff
@@ -0,0 +1,351 @@
...
+        }.get(period.lower())
+
+        if period_seconds is None:
+            logger.warning("Invalid throttle period: %s", period)
+            return None
+
+        return (limit, period.lower(), period_seconds)
+    except (ValueError, AttributeError) as e:
+        logger.warning("Invalid throttle rate format '%s': %s", rate, e)
+        return None
```
jbpenrath: WDYT to implement a custom value `RateValue` (Like we did with `JSONValue`)? class to parse. In this way we would have data validation at application start and do not to parse value each time we need it.

src/backend/core/services/throttle.py
```diff
@@ -0,0 +1,351 @@
...
+        # Set expiry only if this is a new key (TTL returns -1 if no expiry set)
+        if client.ttl(cache_key) == -1:
+            client.expire(cache_key, expiry_seconds)
+        return new_value
+
+    # Fallback for non-Redis cache backends (e.g., in tests)
+    current = get_current_usage(cache_key)
+    new_value = current + amount
+    cache.set(cache_key, new_value, expiry_seconds)
+    return new_value
```
jbpenrath: Why not use django cache method?


```suggestion
    if cache.get(cache_key):
        new_value = cache.incr(cache_key, amount)
    else:
        new_value = amount
        cache.set(cache_key, amount, expiry_seconds)
    
     return new_value 
```

src/backend/core/mda/inbound.py
```diff
@@ -20,44 +20,97 @@
...
 
-    try:
-        local_part, domain_name = email_address.split("@", 1)
-    except ValueError:
-        return False  # Invalid format
+    Returns a set of email addresses that are deliverable locally.
+    An email is deliverable if:
+    - MESSAGES_ACCEPT_ALL_EMAILS is True (test mode), or
+    - The domain matches MESSAGES_TESTDOMAIN (catch-all), or
+    - A mailbox exists for that email address
```
jbpenrath, line 71: We could consider an email locally deliverable if the domain of the email exists ? Then we should create the mailbox if it does not exist. Like the previous implementation nope?

src/backend/core/mda/outbound.py
```diff
@@ -68,6 +69,16 @@ def prepare_outbound_message(
...
         )
 
+    # Throttle external recipients per mailbox/maildomain
+    try:
+        check_and_increment_throttle(
+            mailbox=mailbox_sender,
+            maildomain=mailbox_sender.domain,
+            message=message,
+        )
+    except ThrottleLimitExceeded as e:
```
jbpenrath: WDYT to let the `ThrottleLimitExceeded` to raise especially to be able to catch it into the viewset to be able to return a 402 response?

src/backend/core/mda/inbound.py
sylvinus: not ideal creating such mailboxes, if they are wrong nobody will be there to clean them

PR #517, opened 2026-01-29
✨(importer) enhance MBOX file processing with S3 streaming

jbpenrath:
## Purpose

Introduced a new function to stream files directly from S3, avoiding memory issues with large MBOX files. The processing now reads messages in a single pass, improving efficiency and maintaining message order for threading.

## Summary by CodeRabbit

* **New Features**
  * Import UI: indeterminate progress state, progress text shows percent or per-message counts; progress API adds hasKnownTotal and currentMessage.
  * Progress bar supports indeterminate mode.

* **Bug Fixes**
  * MBOX import streams and indexes remote files, fetches byte ranges to reduce memory; two-pass indexing/processing with finer per-message progress and success/failure tracking.

* **Documentation**
  * Added English and French pluralized strings for imported message counts.

* **Tests**
  * Updated tests for streaming/indexing, date extraction, ordering, threading, and new progress semantics.

### DIFF
--- a/src/backend/core/services/importer/tasks.py
+++ b/src/backend/core/services/importer/tasks.py
@@ -1,7 +1,10 @@
 """Import-related tasks."""
 
 # pylint: disable=unused-argument, broad-exception-raised, broad-exception-caught, too-many-lines
-from typing import Any, Dict, Generator
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from email.parser import BytesParser
+from typing import Any, Dict, List, Optional
 
 from django.core.files.storage import storages
 
@@ -11,6 +14,7 @@ from celery.utils.log import get_task_logger
 from core import enums
 from core.mda.inbound import deliver_inbound_message
 from core.mda.rfc5322 import parse_email_message
+from core.mda.rfc5322.parser import parse_date
 from core.models import Mailbox
 
 from messages.celery_app import app as celery_app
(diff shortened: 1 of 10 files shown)


src/backend/core/services/importer/tasks.py
```diff
@@ -27,6 +31,238 @@
...
+    Stores byte offsets and date for sorting messages chronologically
+    without loading full message content into memory.
+    """
+
+    start_byte: int
+    end_byte: int
+    date: Optional[datetime]
+
+
+def extract_date_from_headers(raw_message: bytes) -> Optional[datetime]:
```
sylvinus, line 47: do we absolutely need this? This seems potentially error-prone as it introduces a second path for parsing emails.

sylvinus: Also, the timestamp from the "Received" header, which ~aligns with our created_at, should be prefered rather than the "Date" from the headers?

src/backend/core/services/importer/tasks.py
```diff
@@ -70,70 +306,121 @@ def process_mbox_file_task(self, file_key: str, recipient_id: str) -> Dict[str,
...
+                    "failure_count": 0,
+                    "type": "mbox",
+                    "current_message": 0,
+                },
+                "error": None,
+            },
+        )
+
+        # ===== PASS 2: Process messages in chronological order =====
+        for idx, msg_index in enumerate(message_indices):
```
sylvinus, line 379: could we batch N messages here, with conditions: 1) the next one is contiguous (actually prepended) 2) the total bytes to fetch doesn't exceed ~100Mo. get_s3_byte_range would fetch the whole batch then use indices to splice

src/backend/core/services/importer/tasks.py
jbpenrath: We could tackle your 2 comments (could we fetch larger range of bytes and can we prevent to rely on date to get message) with a single answer :

We keep the old strategy which assumed that mbox messages are sorted by ascending creation date. In this way the only thing we need is the file byte length then we request byte chunks from the end to the start.

PR #518, opened 2026-01-29
✨(print) add "Print" button in messages context menu

sylvinus:
Capture d’écran 2026-01-29 à 23 11 11

## Summary by CodeRabbit

* **New Features**
  * Added a Print action in thread view to print individual messages from the message actions menu.
  * Print output includes message headers (From, To, Cc, Bcc, Date, Subject) and the message body, formatted for printing and auto-opens the print dialog.
  * Added UI translations for the "Print" label in English, French, and Dutch.

### DIFF
--- a/src/frontend/src/features/message/use-print.tsx
+++ b/src/frontend/src/features/message/use-print.tsx
@@ -1,0 +1,77 @@
+import { useCallback } from "react";
+import { renderToStaticMarkup } from "react-dom/server";
+import { useTranslation } from "react-i18next";
+import { Message } from "@/features/api/gen/models";
+
+const formatContact = (c: { name?: string | null; email: string }) =>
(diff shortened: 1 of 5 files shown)


src/frontend/src/features/message/use-print.tsx
```diff
@@ -0,0 +1,78 @@
...
+    c.name ? `${c.name} <${c.email}>` : c.email;
+
+/**
+ * Hook to print a message in a new browser window
+ */
+const usePrint = () => {
+    const { t, i18n } = useTranslation();
+
+    const print = useCallback((message: Message) => {
+        const iframe = document
```
jbpenrath, line 16: We could return if no iframe content is found instead to print an empty doc

PR #532, opened 2026-02-04
🐛(backend) delete orphan attachments when removed from draft

jbpenrath:
## Purpose

When users remove attachments from drafts, only the M2M relationship was being severed, leaving Attachment and Blob records orphaned in the database. Over time, this leads to unnecessary storage consumption and data accumulation.

A management command is also provided to clean up existing orphans.

## Summary by CodeRabbit

* **New Features**
  * Management command to find and delete orphan attachments and their blobs, with domain/mailbox filtering and a dry-run preview.

* **Improvements**
  * Draft update flow now removes orphaned attachment and blob records when attachments are detached.

* **Tests**
  * Added a test verifying orphan attachments and blobs are removed when a draft’s attachment is deleted.

### DIFF
--- a/src/backend/core/management/commands/delete_orphan_attachments.py
+++ b/src/backend/core/management/commands/delete_orphan_attachments.py
@@ -1,0 +1,121 @@
+"""Delete orphan attachments that are not linked to any message."""
+
+from django.core.management.base import BaseCommand, CommandError
+
+from core import models
+
+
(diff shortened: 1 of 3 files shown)


sylvinus (approved):
IMHO we should target a future migration of models.Attachment to models.DraftAttachment which wouldn't be m2m

PR #566, opened 2026-02-25
✨(threadevents) add new ThreadEvent model

sylvinus:
For now we implement type=im for private messaging among team members of a shared inbox. 

Reimplements part of #383 

https://github.com/user-attachments/assets/1f829cbe-db1e-425d-9311-a826bed8e48d

## Release Notes

* **New Features**
  * Added internal instant messaging (IM) within threads with @mention support and user autocomplete
  * Added message editing with visual "edited" indicators
  * Added message deletion with confirmation dialogs
  * Added thread member list visibility

* **Improvements**
  * Enhanced thread view layout responsiveness
  * Improved link styling consistency across components

* **Localization**
  * Added translations for messaging features (English & French)


src/frontend/public/locales/common/it-IT.json
jbpenrath, line 1: Here it is my instruction I put into Claude.md to prevent it to translate all and bypassing i18next-cli.

```
## Internationalization

You should never modify translation files manually. You should use the `make i18n-generate` command to generate the translation files.

Once the files are updated, you can complete the empty translations. You will only translate French and English for now. You will revert changes on other languages.
```

PR #641, opened 2026-04-20
Perf/search indexation

jbpenrath:
## Purpose

Opensearch index is updated each time a Thread, Message or MessageRecipient
is updated through signals. The current logic has performance issues has
n update of a resource will generate n celery task to update the same
resource... So this work aims to batch updates. Index is updated each
30s and resource ids is deduplicated to only update a resource once.
Furthermore, in an import context, the index will be updated only
when the import will be completed to prevent to spam the celery
worker with numerous indexation task.

`bulk_data` method was missing some options (timeout, max_retries) so 
sometimes when the payload was heavy, the request can failed because 
the timeout was too slow by default (10s) and there is no `max_retries`. 
Furthermore, we build bulk payload by chunking thread but it did not 
check the payload bytes size, so in case of heavy message, the payload 
could be heavy. Now a max_bytes limit is set (50 Mib by default)

## Summary by CodeRabbit

* **New Features**
  * Coalesced/batched thread reindexing with bulk reindex and bulk delete tasks, plus a periodic drain job.

* **Configuration**
  * Added OPENSEARCH_BULK_TIMEOUT, OPENSEARCH_BULK_MAX_BYTES, and SEARCH_REINDEX_TASKS_INTERVAL; worker memory limit set to 1GB.

* **Behavior**
  * Signal and import paths defer and coalesce reindex/delete work, batching side effects to reduce redundant OpenSearch operations.

* **Documentation**
  * Added search/indexation docs and clarified OpenSearch timeout/bulk settings.

* **Tests**
  * Expanded tests for deferral, coalescing, batching, transactional correctness, and bulk error handling.

### DIFF
--- a/docs/env.md
+++ b/docs/env.md
@@ -71,7 +71,9 @@ The application uses a new environment file structure with `.defaults` and `.loc
 | Variable | Default | Description | Required |
 |----------|---------|-------------|----------|
 | `OPENSEARCH_URL` | `["http://opensearch:9200"]` | OpenSearch hosts list | Optional |
-| `OPENSEARCH_TIMEOUT` | `20` | OpenSearch query timeout | Optional |
+| `OPENSEARCH_TIMEOUT` | `20` | OpenSearch query timeout (seconds) for unitary requests | Optional |
+| `OPENSEARCH_BULK_TIMEOUT` | `60` | OpenSearch request timeout (seconds) applied to bulk indexation calls. Raise it if full reindex (`make search-index`) hits timeouts on large payloads. | Optional |
+| `OPENSEARCH_BULK_MAX_BYTES` | `52428800` | Flush threshold (bytes) for bulk indexation payloads; default 50 MiB. Once accumulated actions exceed this, `opensearch-py` emits a sub-chunk HTTP request. Note: this is a batching threshold, not a per-document cap — a single oversized document is still sent as its own chunk. Keep well under the OpenSearch server `http.max_content_length` | Optional |
 | `OPENSEARCH_INDEX_THREADS` | `True` | Enable thread indexing | Optional |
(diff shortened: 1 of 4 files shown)


src/backend/core/services/search/tasks.py
```diff
@@ -210,10 +229,99 @@ def index_message_task(self, message_id):
...
+    create_index_if_not_exists()
+
+    result = _bulk_reindex_threads(models.Thread.objects.filter(id__in=thread_ids))
+
+    return {
+        "success": True,
+        "total": result["total"],
+        "success_count": result["indexed_threads"],
+        "failure_count": result["failure_count"],
+    }
```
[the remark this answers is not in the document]
jbpenrath: There is already a retry logic on the Opensearch request. IMO, if we reach the retry limit from Opensearch, it is not necessary to retry the task and block the worker.

sylvinus (approved):
great work

docs/search-indexation.md
```diff
@@ -0,0 +1,346 @@
...
+
+All index writes happen asynchronously through Celery on the `reindex` queue.
+A full reindex is available through a management command for recovery and
+index-schema changes.
+
+## Index Model
+
+The index is named `messages` (see
+`src/backend/core/services/search/mapping.py`). It uses an OpenSearch
+**parent-child join** so a `Thread` document is the parent of its `Message`
```
chibenwa, line 29: I'm curious to know how well this perform on prod. 
In Apache James we kept clear of parent-child relationship for fear for for bad search-time performance.
I'm thus interested if you have benchmarks available for the different mappings on realistic production-like  workload to back this choice.
Regards,
Benoit

jbpenrath: Unfortunately nope, since the beginning of this project, we always link thread and message through this kind of relationship in the index but thanks for your comment, I'll keep this in mind.

PR #660, opened 2026-05-06
🌐(frontend) localized attachment separator

jbpenrath:
## Purpose

The `ATTACHMENT_SEPARATORS` only contains a english separator and as this
string is displayed within other mail client, for non-english users it can
be weird to see an english string. That's why we add new separators and according to the sender language, we use the localized separator that correspond to him.

## Release Notes

* **New Features**
  * Attachment separators in email messages now support multiple languages including English, French, and Dutch. The system automatically falls back to English for unsupported languages and correctly extracts attachments regardless of the separator language used.

* **Tests**
  * Added comprehensive test coverage for localized attachment separator functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/suitenumerique/messages/pull/660)

### DIFF
--- a/src/frontend/src/features/utils/mail-helper.tsx
+++ b/src/frontend/src/features/utils/mail-helper.tsx
@@ -1,6 +1,7 @@
 import { renderToString, renderToStaticMarkup } from "react-dom/server";
 import { Markdown } from "@react-email/components";
 import DetectionMap from "@/features/i18n/attachments-detection-map.json";
+import i18n from "@/features/i18n/initI18n";
 import z from "zod";
 import { DriveFile } from "../forms/components/message-form/drive-attachment-picker";
 import { handle } from "./errors";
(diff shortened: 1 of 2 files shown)


src/frontend/src/features/utils/mail-helper.tsx
```diff
@@ -42,8 +43,22 @@ export const SUPPORTED_IMAP_DOMAINS = new Map<string, ImapConfig>([
...
+export const ATTACHMENT_SEPARATORS = [
+    '---------- Drive attachments ----------',
+    '---------- Fichiers joints ----------',
+    '---------- Drive-bijlagen ----------',
+];
+
+// Active separator used when sending a new message, keyed by i18n language code.
+// Unknown languages fall back to the legacy English value (first entry above).
+const ATTACHMENT_SEPARATORS_BY_LANG: Record<string, string> = {
+    'en-US': '---------- Drive attachments ----------',
```
sylvinus, line 55: why aren't those in the translations system ?

jbpenrath: For backward compatibility, we cannot remove a separator entry, we have to add new ones. So managing those keys through translation systems introduces, according to me, a high risk of key removal.

sylvinus (approved):

PR #691, opened 2026-06-03
✨(integrations) add outgoing webhooks for every incoming email

sylvinus:
## Summary by CodeRabbit

* **New Features**
  * Added webhook integrations for outbound message delivery, including new trigger options, payload formats, and credential handling.
  * Added support for capturing richer message and event details, including webhook actor display names and one-time secret rotation.

* **Bug Fixes**
  * Improved inbound delivery handling, including safer retries, abandoned message recovery, and better spam/safety warnings.
  * Added internal-delivery support for same-instance mail and clearer delivery-status handling.

* **Documentation**
  * Expanded webhook, environment, and tiered-storage documentation.
  * Updated localization text for the new webhook and delivery UI.


sylvinus:
@coderabbitai review

sylvinus:
@coderabbitai review

src/backend/core/management/commands/backfill_postmark.py
jbpenrath, line 1: Message model has a `post_save` signal which trigger a thread reindexation. That means that at each save call here, one opensearch request will be done... furthermore to update data that opensearch doesn't care.

Maybe we could use `bulk_update` ?

src/backend/core/management/commands/backfill_postmark.py
```diff
@@ -0,0 +1,128 @@
...
+
+        scanned = 0
+        populated = 0
+        errors = 0
+
+        while scanned < limit:
+            take = min(batch_size, limit - scanned)
+            # Processed rows leave the ``postmark__isnull=True`` set, so the next
+            # slice is always fresh work — no cursor/offset needed.
+            batch = list(qs.order_by("created_at")[:take])
```
jbpenrath: In dry mode this will not work as the qs content does not change as we don't populate postmark field. We never reach `if not batch` statement so if there is only one message to update, we will loop on the same message until scanned equals to limit.

PR #692, opened 2026-06-03
✨(mta-in) rewrite MTA-in in pure Python to remove dep on Postfix

sylvinus:
Postfix was already removed as a mta-out dependency, this is the second step so we have a pure python, more auditable path for incoming emails. We plan to keep postfix as a compatible option for a while but it won't be the default once this is battle tested.

## Summary by CodeRabbit

* **New Features**
  * Added a pure-Python inbound SMTP server option (pymta) alongside the existing Postfix-based implementation, with hardened container/runtime variants (including distroless) and dedicated test services.
  * Added Prometheus metrics support (and metrics integration tests) for the pure-Python implementation.

* **Documentation**
  * Updated inbound MTA README with both implementations, ports, switching guidance, and pymta-specific environment variables.

* **Bug Fixes**
  * Improved JWT creation to avoid claim overwrites and use UTC-based expiration.

* **Tests**
  * Added unit, security-hardening, metrics, and limit/MDA client tests; updated test setup to be env-driven, with compose/Makefile targets for both implementations.


sylvinus:
@coderabbitai review

PR #711, opened 2026-06-15
🔧(deps) install jmap-email from pypi

jbpenrath:
## Purpose

Now that jmap-email 0.1.0 is available on pypi we install it from this registry and remove all tweaks to install the deps from local folder. We keep the volume override for backend services in order to be able to work on jmap-email and test it with ease in local development environment.

## Summary by CodeRabbit

* **Chores**
  * Simplified Docker build configuration and CI/CD workflows by removing unused context dependencies.
  * Updated the jmap-email dependency to a pinned version 0.1.0 for more predictable and reproducible builds.


sylvinus (approved):
Better!

PR #721, opened 2026-06-24
!wip(import) resumable import orchestration

jbpenrath:
## Purpose

Work in progress on resumable/cancellable message import: introduce an orchestrator and per-channel import service so imports can be tracked, resumed and cancelled, with supporting API, models and frontend status polling.


src/frontend/src/features/forms/components/message-form/index.tsx
```diff
@@ -792,7 +790,7 @@ export const MessageForm = forwardRef<MessageFormHandle, MessageFormProps>(({
                     <MessageComposer
                         ref={composerRef}
                         mailboxId={form.getValues('from')}
-                        defaultValue={form.getValues('messageDraftBody')}
```
sylvinus, line 795: @jbpenrath is this linked to imports?

jbpenrath: nope

PR #726, opened 2026-06-25
Clean/html body generate

jbpenrath:
## Purpose

We miss to add X-Mailer header to our outbound messages and this
missing can increase the spam score of our messages.

We are currently using react-email to generate html bodies. This
library aims to generate marketing email consistent in all mail
providers. For personal message, it generates too much custom styles
that can increase spam score of those messages.

## Summary by CodeRabbit

* **New Features**
  * Outgoing emails now include an `X-Mailer` header with the product name, automatically appending the release version when available (and preserving a caller-supplied value).
* **Bug Fixes**
  * Improved exported email HTML with cleaner native markup and tighter link/image output (safe links only, unsafe schemes dropped while keeping text, and better image layout/spacing).
* **Refactor**
  * Removed the markdown-to-HTML conversion utility and updated the mail helper tests accordingly.
* **Tests**
  * Added coverage validating `X-Mailer` behavior on both standard and raw MIME submission paths.


src/backend/core/mda/outbound.py
```diff
@@ -113,6 +113,25 @@ def validate_attachments_size(total_size: int, message_id: str) -> None:
 # typically Bcc-only — looking legitimate without disclosing anyone.
 UNDISCLOSED_RECIPIENTS_TO_HEADER = b"To: undisclosed-recipients:;"
 
+# Product name advertised in the outbound X-Mailer header (the running release
+# is appended). See compose_and_sign_mime.
+XMAILER_PRODUCT = "Messages"
```
sylvinus: this should be a settings value

src/backend/core/mda/outbound.py
```diff
@@ -113,6 +113,25 @@ def validate_attachments_size(total_size: int, message_id: str) -> None:
...
+def build_xmailer_value() -> str:
+    """Return the X-Mailer header value: product name plus running release.
+
+    A present X-Mailer is a small positive deliverability signal (its absence
+    reads as bulk/templated mail to some filters, e.g. iCloud). Like other MUAs
+    (e.g. Open-Xchange's "Open-Xchange Mailer vX.Y.Z-RevN"), this identifies the
+    software, not the deployment — so the product name is fixed and the running
+    release appended when available.
+    """
+    release = getattr(settings, "RELEASE", None)
```
sylvinus: avoid `getattr(settings`

PR #733, opened 2026-06-30
🐛(mta-out) de-indent relay block in main.cf.j2 so relay SASL auth works

corentingiraud:
## Problem

The relay section of `src/mta-out/etc/main.cf.j2` is indented by one space:

```jinja
relayhost = {{ SMTP_RELAY_HOST }}
 {% if SMTP_RELAY_USERNAME and SMTP_RELAY_PASSWORD %}
 smtp_sasl_auth_enable = yes
 smtp_sasl_password_maps = hash:/etc/postfix/sasl/relay_passwd
 smtp_sasl_security_options = noanonymous
 {% else %}
 ...
```

Jinja keeps that leading whitespace in the rendered `/etc/postfix/main.cf`, and **Postfix treats any line beginning with whitespace as a continuation of the previous parameter**. So the `smtp_sasl_*` directives get folded into the `relayhost` value instead of being set as parameters. `postconf` confirms it:

```
relayhost = [relay.example.com]:587 smtp_sasl_auth_enable = yes smtp_sasl_password_maps = hash:/etc/postfix/sasl/relay_passwd smtp_sasl_security_options = noanonymous
smtp_sasl_auth_enable = no      # ← stays default; the line above was swallowed
```

Consequences when relaying through an authenticated upstream (e.g. a transactional-email provider):

- `smtp_sasl_auth_enable` is never enabled, so Postfix sends `MAIL FROM` without authenticating → the relay rejects with `502 5.7.0 Please authenticate first`.
- the malformed `relayhost` also triggers `fatal: valid hostname or network address required in server description: =`.

Outbound mail silently fails (messages defer/bounce).

## Fix

De-indent the relay block to column 0 so each directive is parsed as its own parameter. No behavioural change other than the directives now taking effect.

## Verification

Built from this branch and pointed at an authenticated relay (Scaleway TEM):

- before: `502 Please authenticate first`, `postconf` shows `smtp_sasl_auth_enable = no`
- after: `postconf` shows `smtp_sasl_auth_enable = yes`; delivery succeeds (`status=sent (250 ... OK: queued ...)`)

## Summary by CodeRabbit

* **Improvements**
  * Enhanced email relay configuration with conditional upstream authentication when a relay host is set.
  * Relay authentication now enables only when both username and password are provided; otherwise it remains disabled.
  * Uses a relay password map and disallows anonymous authentication for improved security.
* **Bug Fixes**
  * Fixed an issue where relay block indentation could break SASL authentication behavior.


sylvinus:
thanks!

PR #750, opened 2026-07-16
📦(frontend) setup browserlist and support chrome >= 109

jbpenrath:
## Purpose

Some of our users has reported issues using the application with Chrome 109. Instead of polyfilling one by one each method, we setup vite legacy plugin and configured a browserlist. It adds a 23.5Kb Gzipped module but the polyfill strategy management is more standard and robust.

Then with the upgrade to vite 8.1.5 we save ~400kb (~125kb gzipped) on the global bundle size 😉

Close #741

## Summary by CodeRabbit

* **New Features**
  * Improved browser compatibility via automatic runtime polyfills for modern builds, aligned to the supported Chrome/Edge baseline.

* **Improvements**
  * Refreshed the frontend build tooling and browser targeting rules to match the supported audience.
  * Updated build minification for faster, more efficient production output.

* **Chores**
  * Updated development dependencies to newer Vite tooling.

### DIFF
--- a/src/frontend/package-lock.json
+++ b/src/frontend/package-lock.json
@@ -48,6 +48,7 @@
         "@types/node": "25.2.0",
         "@types/react": "19.2.10",
         "@types/react-dom": "19.2.3",
+        "@vitejs/plugin-legacy": "8.2.1",
         "@vitejs/plugin-react": "6.0.2",
         "@vitest/coverage-v8": "4.1.8",
         "@vitest/ui": "4.1.8",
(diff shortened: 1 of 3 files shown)


jbpenrath:
@sylvinus I know that keeping bundle size as small as possible is important for you. So before merging that, I would like to know if you are ok to increase a little bit the bundle size (~66kb / 23.5kb gziped) in favor of maintanability and support of some "old" browsers.

sylvinus:
@jbpenrath ok for me. Shouldn't we also add a target minimum firefox version? I think @lebaudantoine knows which version some ministries are pinned at, which could be a good floor?

sylvinus:
Also do we have a clean message in case of non-support for older browsers?

PR #751, opened 2026-07-18
✨(frontend+backend) show per-thread message count badge in mailbox list

nicolasaunai:
## Purpose

Adds a per-thread email-count badge in the mailbox list so users can see at a glance how many messages a thread contains.

## Proposal

- Backend: `ThreadSerializer` gains a `message_count` field, annotated as `Count("messages", filter=Q(messages__is_draft=False), distinct=True)` on the thread list queryset to avoid JOIN multiplication from the accesses join.
- Frontend: a neutral `Badge` renders the count (only when > 1) in the thread list item, with an i18n aria-label/title.
- Tests: `TestThreadListMessageCount` covers zero, exact count, draft exclusion, and distinct-per-thread regression.

- [x] signoff
- [x] signed commits
- [x] lint
- [x] tests

## Summary by CodeRabbit

* **New Features**
  * Added a message count badge to mailbox thread entries when a thread contains multiple messages.
  * Message counts exclude draft messages and are calculated separately for each thread.

* **Documentation**
  * Updated the changelog to document the new thread message count badge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to
 - Bump keycloak to 26.6.3
 - Bump keycloak to 26.6.4
 
+### Added
+
+- ✨(frontend+backend) show per-thread message count badge in mailbox list
+
 ### Fixed
 
 - 🐛(mta-out) fix relay block indentation breaking SASL auth #733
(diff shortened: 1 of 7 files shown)


jbpenrath:
Done in #755

image

PR #756, opened 2026-07-20
✨(inbound) ARC relay-trust: inbound_auth "arc" + arc_gate

mosa-riel:
## What
Add ARC-based sender trust for inbound mail, in `SPAM_CONFIG`:
- **`inbound_auth: "arc"`** — verdict backend that reads dkim/dmarc **only** from an `ARC-Authentication-Results` sealed by a trusted sealer. Plaintext headers are never read.
- **`arc_gate`** — enforce that mail is sealed by a trusted relay: `off | spam | drop`.

## Why
With `inbound_auth: authentication-results`, dkim/dmarc come from the top-level `Authentication-Results`, trusted by Received-block position. A host delivering straight to the MDA can forge that header (and `From:`) and show as verified. When inbound is relayed by a provider that ARC-seals (a filtering relay in front of a public MX), the seal is an unforgeable anchor — verify it instead of trusting a plaintext header.

## Config
```jsonc
{
  "inbound_auth": "arc",
  "trusted_arc_sealers": ["relay.example"],  // empty = any valid seal
  "arc_gate": "spam"                          // off | spam | drop
}
```
- `arc` mode → verdict from the trusted sealer's sealed AAR; unsealed → unverified.
- `arc_gate` → action when not trusted-sealed: `spam` (`is_spam=True`) / `drop` (discard).

## Design choices
- **Sealer allowlist, not just "valid ARC".** A valid chain only proves *some* domain sealed it; an attacker self-seals. The outermost sealer (highest `i=`) must be allowlisted. Empty list = any valid seal (weak; only safe when the MDA is IP-restricted to the relay).
- **No plaintext-header fallback in `arc` mode** — falling back reopens the hole.
- **DNS/verify failure never spams or drops** — can't-verify ≠ forged.
- **Gate runs first among spam steps** so untrusted is authoritative.
- **Kept in `SPAM_CONFIG`** beside `inbound_auth`, reusing per-domain overrides. (These keys are auth/trust, not spam — a split is possible but out of scope.)

## Tests
`core/mda/arc.py` verifier, `arc` verdict mode, gate step — incl. a hermetic real-crypto case (synthetic seal + stub DNS). No new dependency (dkimpy already vendored). Config reference: `docs/spam-config.md`.

## Follow-up (PR2)
- `arc_gate: quarantine` — hold untrusted mail (recoverable, admin-visible) via a new `InboundMessage` state — a first-class "untrusted" bucket instead of reusing spam/drop.
- Tighten `dnsfail`: separate transient DNS from malformed/hostile input.

## Summary by CodeRabbit

* **New Features**
  * Added ARC-based sender authentication for inbound messages.
  * Added `trusted_arc_sealers` configuration and `arc_gate` actions to allow, mark as spam, or drop based on verified ARC sealing.
  * Ensures forged or untrusted top-level authentication headers are ignored when ARC trust isn’t established.
* **Documentation**
  * Expanded `SPAM_CONFIG` documentation with detailed environment-variable semantics, examples, and security guidance.
* **Tests**
  * Added coverage for ARC sealer trust matching, ARC verdict derivation, and `arc_gate` pipeline behavior.

### DIFF
--- a/docs/spam-config.md
+++ b/docs/spam-config.md
@@ -1,0 +1,104 @@
+# `SPAM_CONFIG`
+
+Inbound spam filtering **and** sender-authentication config. Set globally via the
+`SPAM_CONFIG` env var (JSON), overridable per mail domain through
+`MailDomain.custom_settings["SPAM_CONFIG"]` (merged by
+`MailDomain.get_spam_config()`).
+
+> Naming note: these keys cover both spam scoring and sender auth/trust. The
+> auth keys live here because `inbound_auth` already did — not because they are
+> "spam". A future split into a dedicated auth/trust config is possible but out
+> of scope here.
+
+## Keys
+
+| Key | Type | Default | Purpose |
+| --- | --- | --- | --- |
+| `rspamd_url` | string | — | rspamd `/checkv2` endpoint. Absent = no rspamd scan. |
+| `rspamd_auth` | string | — | Optional `Authorization` header value for rspamd. |
+| `rules` | list | `[]` | Hardcoded header-match spam rules (see below). |
+| `trusted_relays` | int | `0` | How many upstream `Received` blocks to trust for the header-match `rules`. Block 0 = the block our own MTA prepends. Not used by `inbound_auth: "arc"`. |
+| `inbound_auth` | string\|null | `null` | Sender-auth backend for the verdict/banner (see below). |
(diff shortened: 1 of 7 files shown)


sylvinus:
Thanks!! merged as https://github.com/suitenumerique/messages/commit/cf0b70e7eefa59a5537c06817ac17a015d6749b1

